PR #18, opened 2015-05-06
Add support to prism to output a particular field

sihil:
This allows you to get the IP address, AMI or other field instead of dnsName if it is more appropriate. Particularly useful if you are working on hosts in a VPC for which DNS resolution is unavailable.


theefer:
Nice :+1:

PR #22, opened 2015-12-16
Modernisation of Prism for AWS hosting

sihil:
Prism is the first part of moving our deployment and discovery stack into AWS.

Some decisions that we are making:
 - We will orphan the new prism installation (i.e. the new Prism will only know about AWS instances and nothing about GC2)
 - Prism should start to use AssumeRole rather than possessing credentials for all accounts

Apologies for the large PR.

 - [x] Update SBT, Scala, Play, Akka etc.
 - [x] Remove Openstack support
 - [x] Convert to use `sbt-riffraff-artifact`
 - [x] Change to use published `guardian-management-play` library when released
 - [x] Create a cloudformation file
 - [x] Change the auth mechanism to use AssumeRole instead of Basic credentials


sihil:
I've created a branch of the existing master called GC2 for the old codebase.

jennysivapalan:
Nice work @sihil! :star:

PR #25, opened 2016-02-16
Fix case where addresses object doesn't exist

sihil:


sihil:
Pushed version 0.7.1.

PR #28, opened 2016-02-22
Add launch configurations to the the items Prism indexes

sihil:
This adds launch configurations to the list of items that Prism indexes from each account.

Sadly prism doesn't have permission to call this at present so we must update the cloudformation in each account.

### DIFF
--- a/app/collectors/launchConfigurations.scala
+++ b/app/collectors/launchConfigurations.scala
@@ -13,11 +13,11 @@ import scala.util.Try
 
 object LaunchConfigurationCollectorSet extends CollectorSet[LaunchConfiguration](ResourceType("launch-configurations", Duration.standardMinutes(15L))) {
   val lookupCollector: PartialFunction[Origin, Collector[LaunchConfiguration]] = {
-    case amazon:AmazonOrigin => AWSLaunchConfigurationCollector(amazon, resource)
+    case amazon: AmazonOrigin => AWSLaunchConfigurationCollector(amazon, resource)
   }
 }
 
-case class AWSLaunchConfigurationCollector(origin:AmazonOrigin, resource:ResourceType) extends Collector[LaunchConfiguration] with Logging {
+case class AWSLaunchConfigurationCollector(origin: AmazonOrigin, resource: ResourceType) extends Collector[LaunchConfiguration] with Logging {
 
   val client = new AmazonAutoScalingClient(origin.credentials.provider)
   client.setRegion(origin.awsRegion)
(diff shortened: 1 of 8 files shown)


app/collectors/launchConfigurations.scala
```diff
@@ -0,0 +1,70 @@
...
+      keyName = config.getKeyName,
+      placementTenancy = Option(config.getPlacementTenancy),
+      securityGroups = Option(config.getSecurityGroups).map(_.asScala).toList.flatten,
+      userData = Option(config.getUserData)
+    )
+  }
+}
+
+case class LaunchConfiguration(
+                  arn: String,
```
adamnfish: You should probably update your IntelliJ settings so it lays out multi-line case classes properly. Although if everything else in this repo is like that maybe leave it for consistency, up to you.

Settings -> Editor -> Code-style -> Scala -> Wrapping and Braces. Disable "align when multiline" and enable "use normal indent for parameters".

[Evidence that this is correct](http://docs.scala-lang.org/style/indentation.html).

app/collectors/launchConfigurations.scala
```diff
@@ -0,0 +1,70 @@
...
+      arn = config.getLaunchConfigurationARN,
+      name = config.getLaunchConfigurationName,
+      imageId = config.getImageId,
+      region = regionName,
+      instanceProfile = Option(config.getIamInstanceProfile),
+      createdTime = Try(new DateTime(config.getCreatedTime)).toOption,
+      instanceType = config.getInstanceType,
+      keyName = config.getKeyName,
+      placementTenancy = Option(config.getPlacementTenancy),
+      securityGroups = Option(config.getSecurityGroups).map(_.asScala).toList.flatten,
```
adamnfish: If you import `import collection.JavaConversions._` I don't think you'd need that `.map(_asScala)`?

app/collectors/launchConfigurations.scala
```diff
@@ -0,0 +1,70 @@
...
+
+case class AWSLaunchConfigurationCollector(origin:AmazonOrigin, resource:ResourceType) extends Collector[LaunchConfiguration] with Logging {
+
+  val client = new AmazonAutoScalingClient(origin.credentials.provider)
+  client.setRegion(origin.awsRegion)
+
+  def crawlWithToken(nextToken: Option[String]): Iterable[LaunchConfiguration] = {
+    val request = new DescribeLaunchConfigurationsRequest().withNextToken(nextToken.orNull)
+    val result = client.describeLaunchConfigurations(request)
+    val configs = result.getLaunchConfigurations.asScala.map { LaunchConfiguration.fromApiData(_, origin.region) }
```
adamnfish: Again, I think better than `asScala` is importing the `JavaConversions`.

app/collectors/launchConfigurations.scala
sihil: `JavaConvertors` is the new `JavaConversions`, which has been deprecated in preference of being explicit.

app/collectors/launchConfigurations.scala
sihil: As above, `JavaConversions` is deprecated.

app/collectors/launchConfigurations.scala
sihil: Done.

app/collectors/launchConfigurations.scala
adamnfish: Good to know, thank you!

PR #30, opened 2016-02-26
Index server certificates

cb372:
Add a collector that crawls all server certificates stored in IAM.

This will allow us to build tooling to check for expiring certificates.


cb372:
Ping @sihil

PR #35, opened 2016-10-05
Add s3 buckets to the the items Prism indexes

### LINKED ISSUE #33: Add collector for S3 bucket
Add the ability to find by which AWS account a S3 bucket is owned.

mchv:
This adds s3 buckets to the list of items that Prism indexes from each account.

Sadly prism doesn't have permission to call this at present so we must update the cloudformation in each account.

Added a commit to update to scala `2.11.8` as well.

Fixes #33 

@sihil

### DIFF
--- a/app/collectors/bucket.scala
+++ b/app/collectors/bucket.scala
@@ -1,0 +1,63 @@
+package collectors
+
+import agent._
+import com.amazonaws.services.s3.AmazonS3Client
+import controllers.routes
+import org.joda.time.{DateTime, Duration}
+import play.api.mvc.Call
+import utils.Logging
+import collection.JavaConverters._
+import com.amazonaws.services.s3.model.{Bucket => AWSBucket, ListBucketsRequest}
+
+import scala.util.Try
+
+object BucketCollectorSet extends CollectorSet[Bucket](ResourceType("bucket", Duration.standardMinutes(15L))) {
+  val lookupCollector: PartialFunction[Origin, Collector[Bucket]] = {
+    case amazon: AmazonOrigin => AWSBucketCollector(amazon, resource)
+  }
+}
+
+case class AWSBucketCollector(origin: AmazonOrigin, resource: ResourceType) extends Collector[Bucket] with Logging {
+
+  val client = new AmazonS3Client(origin.credentials.provider)
+  client.setRegion(origin.awsRegion)
+
+  def crawl: Iterable[Bucket] = {
+    val request = new ListBucketsRequest()
+    client.listBuckets(request).asScala.map {
+      Bucket.fromApiData(_, client)
+    }
+  }
+}
+
+object Bucket {
+
+  private def arn(bucketName: String) = s"arn:aws:s3:::$bucketName" 
+
+  def fromApiData(bucket: AWSBucket, client: AmazonS3Client): Bucket = {
+    val bucketName = bucket.getName
+    Bucket(
+      arn = arn(bucketName),
+      name = bucketName,
+      region = client.getBucketLocation(bucket.getName),
+      createdTime = Try(new DateTime(bucket.getCreationDate)).toOption
+    )
+  }
+}
+
+case class Bucket(
+  arn: String,
+  name: String,
+  region: String,
+  createdTime: Option[DateTime]
+  /*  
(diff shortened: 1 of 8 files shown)


cloudformation/prism-role.template
```diff
@@ -24,7 +24,8 @@
                       "ec2:Describe*",
                       "iam:Get*",
                       "iam:List*",
-                      "autoscaling:Describe*"
+                      "autoscaling:Describe*",
+                      "s3:List*"
```
adamnfish: Looks like `"s3:ListAllMyBuckets"` and `"s3:GetBucketLocation"` are required here. `List*` gives too many permissions and won't be enough for this feature to work.

cloudformation/prism-user.template
```diff
@@ -24,7 +24,8 @@
                       "ec2:Describe*",
                       "iam:Get*",
                       "iam:List*",
-                      "autoscaling:Describe*"
+                      "autoscaling:Describe*",
+                      "s3:List*"
```
adamnfish: (See above)

app/collectors/bucket.scala
```diff
@@ -0,0 +1,63 @@
...
+    )
+  }
+}
+
+case class Bucket(
+  arn: String,
+  name: String,
+  region: String,
+  createdTime: Option[DateTime]
+  /*
```
adamnfish: Does this comment need to be here, or would it be better as an issue on the repo?

app/collectors/bucket.scala
```diff
@@ -0,0 +1,63 @@
...
+  /*  
+      Other properties we may add in the future: 
+
+      Policy                   -> client.getBucketPolicy(String bucketName)
+      ReplicationConfiguration -> client.getBucketReplicationConfiguration(String bucketName)
+      TaggingConfiguration     -> client.getBucketTaggingConfiguration(String bucketName)
+  */
+
+) extends IndexedItem {
+  def callFromArn: (String) => Call = arn => routes.Api.bucket(arn)
```
adamnfish: This is bizzare. I realise it's here from the underlying trait so this is in line with the existing pattern, but this doesn't appear to have anything to do with the instance so it's a strange place to put it. Maybe an `override` keyword here? :this_is_a_minor_point_emoji:

app/collectors/bucket.scala
mchv: I removed it as it was more a reminder for me of things to do in the future. I don't think there is a need now.

philwills:
I think this will be really useful, thanks @mchv

app/collectors/bucket.scala
mchv: I added `override` for clarity.
 Because  `Call` is directly in the model, it kind of make sense to put this mapping here. Otherwise we could have a `mapping` function in the `Api` controller  keep the `play` specific routing isolated :

```scala
def mapping(item: IndexedItem): String => Call = item match {
          case LaunchConfiguration => routes.Api.launchConfiguration
          case Bucket =>  routes.Api.bucket
  }
```

adamnfish:
Awesome work, @mchv. This will be a great help! It's very common to wonder where a bucket lives :-)

mchv:
@adamnfish I have follow current implementation, but the client is not invoked by the async actions but by a scheduled agent. It will be a quite a refactoring to make the agent logic async.

conf/routes
```diff
@@ -20,24 +20,27 @@ GET        /instances/regions             controllers.Api.regionList
...
-GET        /launch-configurations/:arn                    controllers.Api.launchConfiguration(arn)
+GET        /launch-configurations         controllers.Api.launchConfigurationList
+GET        /launch-configurations/:arn    controllers.Api.launchConfiguration(arn)
 
-GET        /server-certificates                        controllers.Api.serverCertificateList
-GET        /server-certificates/:arn                    controllers.Api.serverCertificate(arn)
+GET        /server-certificates           controllers.Api.serverCertificateList
+GET        /server-certificates/:arn      controllers.Api.serverCertificate(arn)
+
+GET        /bucket                        controllers.Api.bucketList
```
sihil: To be consistent can this be `/buckets`?

PR #34, opened 2016-10-05
Adopt new riff-raff.yaml format

sihil:
This is a kludge to make the prism build output a new format build prior to updating the SBT riff-raff plugin.

PR #36, opened 2016-10-05
Update sbt riff-raff plugin and switch to riff-raff.yaml file

sihil:
Supercedes #34 now there is a release of the plugin.

Remove the cloudformation deployment (this should move to the deploy-tools-platform repo).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import scala.language.postfixOps
+
 name := "prism"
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -1,3 +1,5 @@
+import scala.language.postfixOps
```
philwills: Super minor quibble, but why do you need this?

PR #37, opened 2016-10-07
Fetch and list reservation information.

jennysivapalan:
This adds reservation information to the list of items that Prism indexes from each account.

I plan to use this data and the `/instances` endpoint to make decisions on reservations / on-demand for tools accounts.

### DIFF
--- a/app/collectors/reservation.scala
+++ b/app/collectors/reservation.scala
@@ -11,7 +11,7 @@ import utils.Logging
 import collection.JavaConverters._
 import scala.util.Try
 
-object ReservationCollectorSet extends CollectorSet[Reservation](ResourceType("reservations", Duration.standardMinutes(15L))) {
+object ReservationCollectorSet extends CollectorSet[Reservation](ResourceType("reservation", Duration.standardMinutes(15L))) {
   val lookupCollector: PartialFunction[Origin, Collector[Reservation]] = {
     case amazon: AmazonOrigin => AWSReservationCollector(amazon, resource)
   }
--- a/app/collectors/reservation.scala
+++ b/app/collectors/reservation.scala
@@ -2,7 +2,7 @@ package collectors
 
 import agent._
 import com.amazonaws.services.ec2.AmazonEC2Client
-import com.amazonaws.services.ec2.model.{DescribeReservedInstancesRequest, ReservedInstances}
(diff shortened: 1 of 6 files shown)


app/collectors/reservation.scala
```diff
@@ -0,0 +1,56 @@
...
+import com.amazonaws.services.ec2.model.{DescribeReservedInstancesRequest, ReservedInstances}
+import controllers.routes
+import org.joda.time.{DateTime, Duration}
+import play.api.mvc.Call
+import utils.Logging
+
+import collection.JavaConverters._
+import scala.util.Try
+
+object ReservationCollectorSet extends CollectorSet[Reservation](ResourceType("reservations", Duration.standardMinutes(15L))) {
```
sihil: I think I prefer the singular here, but having looked there is no consistency across the code base :(

app/collectors/reservation.scala
```diff
@@ -0,0 +1,56 @@
...
+
+case class Reservation(
+  key: String,
+  region: String,
+  instanceType: String,
+  instanceCount: Int,
+  startTime: Option[DateTime],
+  endTime: Option[DateTime]
+) extends IndexedItem {
+  override def arn: String = s"arn:gu:reservation:key/$key"
```
sihil: Whilst there isn't a real ARN format for reservations I suggest we use something closer to what we imagine it to be. Perhaps:
`arn:aws:ec2:<region>:<accountId>:reservation/<id>`

app/collectors/reservation.scala
jennysivapalan: done!

app/collectors/reservation.scala
jennysivapalan: done!

app/collectors/reservation.scala
```diff
@@ -24,28 +24,30 @@ case class AWSReservationCollector(origin: AmazonOrigin, resource: ResourceType)
...
+  id: String,
+  accountId: Option[String],
   region: String,
   instanceType: String,
   instanceCount: Int,
   startTime: Option[DateTime],
   endTime: Option[DateTime]
 ) extends IndexedItem {
-  override def arn: String = s"arn:gu:reservation:key/$key"
+  override def arn: String = s"arn:aws:ec2:$region:${accountId.getOrElse("")}:reservation/$id"
```
sihil: If you build this in the crawler then you won't need to modify the case class.

jennysivapalan: done and also added the fields as discussed.

PR #38, opened 2016-10-19
Availability zone should be group (as per /instance endpoint) and not region

jennysivapalan:
In https://github.com/guardian/prism/pull/37 I incorrectly put the AZ group as region, I've followed the `/instance` endpoint and renamed it to group.

### DIFF
--- a/app/collectors/reservation.scala
+++ b/app/collectors/reservation.scala
@@ -32,7 +32,7 @@ case class AWSReservationCollector(origin: AmazonOrigin, resource: ResourceType)
 case class Reservation(
   arn: String,
   id: String,
-  group: String,
+  availabilityZone: String,
   instanceType: String,
   instanceCount: Int,
   productDescription: String,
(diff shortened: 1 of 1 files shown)


sihil:
I think `group` might be left over from pre-aws code where AZ didn't make sense everywhere. Suggest we call it `az` or `availabilityZone` rather than group.

sihil:
The ARN should really be region though. Is that available?

jennysivapalan:
Sorry missed the comment..have updated to `availabilityZone`. 

No the region isn't returned. I just reserved some instances that were not tied to an AZ and those reservations do not seem to be returned at all 😞 . The field that determines if a reservation is tied to a region or not is called `Scope` on the AWS console, hopefully it will get added to the library in the future.

sihil:
Have you checked on this recently @jennysivapalan ? Should this be merged regardless?

jennysivapalan:
@sihil oh sorry - I tried to update the library a couple months ago as it did have region and scope in however it caused a fair few compilation issues and I didn't revisit it. We can probably merge this but I'm happy to try again as a Friday type task to see if I can update the library.

PR #40, opened 2017-02-14
Use .deb for deployment

TBonnin:
Modifies prism to use .deb for deployment. This is necessary to move away from machine-images to amigo

Tested in CODE


build.sbt
```diff
@@ -36,16 +38,26 @@ libraryDependencies ++= Seq(
 
 scalacOptions ++= Seq("-feature")
 
+javaOptions in Universal ++= Seq(
+  s"-Dpidfile.path=/dev/null"
```
TBonnin: Related to how Play handles pid file: sbt-native-packager maintainer recommends to let play not handle the pid file.

build.sbt
```diff
@@ -36,16 +38,26 @@ libraryDependencies ++= Seq(
...
+)
+
```
sihil: Does this need @philwills recommended options? See something like https://github.com/guardian/membership-frontend/blob/master/frontend/build.sbt#L60-L68

build.sbt
```diff
@@ -36,16 +38,26 @@ libraryDependencies ++= Seq(
...
 def env(key: String): Option[String] = Option(System.getenv(key))
 
 lazy val root = (project in file("."))
   .enablePlugins(PlayScala, RiffRaffArtifact, UniversalPlugin)
   .settings(
     packageName in Universal := normalizedName.value,
+    maintainer := "Guardian Developers <dig.dev.software@theguardian.com>",
```
sihil: We should create an email address for tools, but this will do for now.

cloudformation/prism.template
```diff
@@ -302,10 +302,13 @@
...
-              "",
+              "\n",
               [
-                "#!/bin/bash -ev\n",
-                "/opt/features/native-packager/install.sh -b deploy-tools-dist -s -t tgz\n"
+                "#!/bin/bash -ev",
+                "mkdir /prism",
+                { "Fn::Join": [ "", [ "aws --region eu-west-1 s3 cp s3://deploy-tools-dist/deploy/", { "Ref": "Stage" }, "/prism/prism.deb /prism/" ] ] },
+                "dpkg -i /prism/prism.deb",
+                "systemctl start prism"
```
sihil: In general I'd love to aim for userdata zero as I think there lies potential for more reusable CFN templates - but that's not enormously relevant here, just my philosophy.

PR #42, opened 2017-02-14
Add ami-cloudformation-parameter deployment

TBonnin:


riff-raff.yaml
```diff
@@ -4,4 +4,14 @@ deployments:
...
+    dependencies:
+      - update-ami
+  update-ami:
+    type: ami-cloudformation-parameter
+    app: prism
+    parameters:
+      amiParameter: MachineImageAMI
+      amiTags:
+        Recipe: xenial-java8-deploy-infrastructure
+        AmigoStage: PROD
```
sihil: I suggest that it would be good practice to put `BuiltBy: amigo` on this list to be confident. What do you think?

TBonnin: @sihil Isn't the `AmigoStage` tag kinda plays the same role of making sure it has been built by amigo?

TBonnin: @sihil added the `BuiltBy: amigo`

sihil: Sorry for not replying to this. For the record we discussed this and my argument was that `AmigoStage` is a proxy for this, but it's not the main point of the tag. In hindsight this should be `Stage` and then we should use `BuiltBy` to determine what platform built it.

I guess the other side of the argument would be that we don't care what builds it, we just want the latest version.

PR #43, opened 2017-02-14
Get riffRaffBuildIdentifier from teamcity env var (not Travis)

TBonnin:
I made a mistake setting the riffRaffBuildIdentifier param. Prism is built on TC, not Travis


build.sbt
```diff
@@ -58,7 +58,7 @@ lazy val root = (project in file("."))
     topLevelDirectory in Universal := Some(normalizedName.value),
     serverLoading in Debian := Systemd,
     riffRaffPackageType := (packageBin in Debian).value,
-    riffRaffBuildIdentifier := env("TRAVIS_BUILD_NUMBER").getOrElse("DEV"),
+    riffRaffBuildIdentifier := env("BUILD_NUMBER").getOrElse("DEV"),
```
sihil: I don't think you should need to set this at all. `sbt-riffraff-artifact` has this logic.
See https://github.com/guardian/sbt-riffraff-artifact/blob/master/src/main/scala/com/gu/riffraff/artifact/RiffRaffArtifact.scala#L71 and https://github.com/guardian/sbt-riffraff-artifact/blob/master/src/main/scala/com/gu/riffraff/artifact/BuildInfo.scala#L68-L80

PR #44, opened 2017-03-29
Add Owners

TBonnin:
This PR is adding the concept of owners into Prism.
Each owner owns a list of SSA (stack, stage, app)

This patch is adding 3 /endpoints:
GET /owners
```

  "status": "success",
  "lastUpdated": "2017-03-29T15:07:13.815+01:00",
  "stale": false,
  "staleSources": [
    
  ],
  "data": {
    "owners": [
      {
        "id": "adam.fisher",
        "stacks": [
          {
            "stack": "security"
          }
        ]
      },
      {
        "id": "dotcom.platform",
        "stacks": [
          {
            "stack": "frontend"
          }
        ]
      },
      {
        "id": "simon.hildrew",
        "stacks": [
          {
            "stack": "deploy"
          }
        ]
      },
      {
        "id": "phil.wills",
        "stacks": [
          {
            
          }
        ]
      }
    ]
  },
  "sources": [
    {
      "resource": "Owners",
      "origin": {
        "vendor": "prism",
        "accountName": "prism"
      },
      "status": "success",
      "createdAt": "2017-03-29T15:07:13.815+01:00",
      "itemCount": 4,
      "age": 0,
      "stale": false
    }
  ]
}
```
            
GET /owners/for-stack?stack=frontend
```
{
  "status": "success",
  "lastUpdated": "2017-03-29T14:00:30.665+01:00",
  "stale": false,
  "staleSources": [
    
  ],
  "data": {
    "owner": [
      {
        "id": "dotcom.platform",
        "stacks": [
          {
            "stack": "frontend"
          }
        ]
      }
    ]
  },
  "sources": [
    {
      "resource": "Owners",
      "origin": {
        "vendor": "prism",
        "accountName": "prism"
      },
      "status": "success",
      "createdAt": "2017-03-29T14:00:30.665+01:00",
      "itemCount": 5,
      "age": 28,
      "stale": false
    }
  ]
}
```

GET /owners/dotcom.platform
```
{
  "status": "success",
  "lastUpdated": "2017-03-29T15:07:13.815+01:00",
  "stale": false,
  "staleSources": [
    
  ],
  "data": {
    "owner": [
      {
        "id": "dotcom.platform",
        "stacks": [
          {
            "stack": "frontend"
          }
        ]
      }
    ]
  },
  "sources": [
    {
      "resource": "Owners",
      "origin": {
        "vendor": "prism",
        "accountName": "prism"
      },
      "status": "success",
      "createdAt": "2017-03-29T15:07:13.815+01:00",
      "itemCount": 4,
      "age": 40,
      "stale": false
    }
  ]
}
```

### DIFF
--- a/app/controllers/OwnerApi.scala
+++ b/app/controllers/OwnerApi.scala
@@ -10,6 +10,7 @@ import play.api.libs.json.Json._
 import play.api.mvc.{Action, RequestHeader, Result}
 import utils.ResourceFilter
 import play.api.http.Status._
+import play.api.mvc.Results._
 
 import scala.concurrent.Future
 
@@ -50,10 +51,12 @@ object OwnerApi {
     val stack: Option[String] = request.getQueryString("stack")
     val stage: Option[String] = request.getQueryString("stage")
     val app: Option[String] = request.getQueryString("app")
-    itemsResult("owner", Owners.forStack(stack, stage, app))
+    val ownerOpt = Owners.forStack(stack, stage, app)
+    val badRequest = BadRequest(s"Invalid request for${stack.fold("")(" Stack: " + _)}${stage.fold("")(" Stage: " + _)}${app.fold("")(" App: " + _)}")
+    ownerOpt.fold(Future.successful(badRequest)) {o => itemsResult("owner", Some(o))}
   }
 
-  def owner(id:String) = Action.async { implicit request =>
+  def owner(id: String) = Action.async { implicit request =>
     val owner = Owners
       .find(id)
       .orElse(throw ApiCallException(Json.obj("id" -> s"Owner with id '$id' doesn't exist"), NOT_FOUND))
--- a/app/controllers/OwnerApi.scala
+++ b/app/controllers/OwnerApi.scala
@@ -1,0 +1,63 @@
+package controllers
+
+import agent.{Label, Origin, ResourceType}
+import data.Owners
+import jsonimplicits.model._
+import model.Owner
+import org.joda.time.{DateTime, Duration}
+import play.api.libs.json.{Json, Writes}
+import play.api.libs.json.Json._
+import play.api.mvc.{Action, RequestHeader, Result}
+import utils.ResourceFilter
+import play.api.http.Status._
+
+import scala.concurrent.Future
+
+object OwnerApi {
+
+  private def itemsResult[T](key: String, items: => Iterable[T])(implicit request: RequestHeader, tw: Writes[T]): Future[Result] = {
+    ApiResult.filter {
+      Map(ownerLabel -> items.toSeq)
+    } reduce { collection =>
+      Json.obj(key -> toJson(collection.values.flatten))
+    }
+  }
+
+  object OwnerResourceType extends ResourceType("Owners", Duration.standardDays(30))
+
+  val ownerLabel = Label(
+    OwnerResourceType,
+    new Origin {
+      val vendor = "prism"
+      val account = "prism"
+      val resources = Set("sources")
+      val jsonFields = Map.empty[String, String]
+    },
+    Owners.all.size,
+    DateTime.now
+  )
+
+  def ownerList = Action.async { implicit request =>
+    val requestFilter =  ResourceFilter.fromRequestWithDefaults()
+    def filter(owner: Owner) = {
+      val json = Json.toJson(owner)
+      if (requestFilter.isMatch(json)) Some(json) else None
+    }
+    itemsResult("owners", Owners.all.toSeq.flatMap(filter))
+  }
+
+  def ownerForStack = Action.async { implicit request =>
+    val stack: Option[String] = request.getQueryString("stack")
+    val stage: Option[String] = request.getQueryString("stage")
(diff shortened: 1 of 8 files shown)


app/controllers/OwnerApi.scala
```diff
@@ -0,0 +1,63 @@
+package controllers
```
TBonnin: I put all the `Owner` endpoints in another controller/file because it is slightly different from the other resources (no agents, crawling, etc...)

app/controllers/OwnerApi.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  private def itemsResult[T](key: String, items: => Iterable[T])(implicit request: RequestHeader, tw: Writes[T]): Future[Result] = {
+    ApiResult.filter {
+      Map(ownerLabel -> items.toSeq)
+    } reduce { collection =>
+      Json.obj("owners" -> toJson(collection.values.flatten))
+    }
+  }
+
+  object OwnerResourceType extends ResourceType("Owners", Duration.standardDays(30))
```
TBonnin: I set`shelfLife` to 30 days but it is quite arbitrary

app/data/Owners.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import model._
+
+trait Owners {
+  def stacks: Set[(String, SSA)]
+
+  def all: Set[Owner] = stacks
+    .groupBy(_._1)
+    .map { case (ownerId, stacks) =>
+      Owner(ownerId, stacks.map(_._2))
```
thomaska: although I understand that the `stacks` here comes from the `case(ownerId, stacks)` maybe for the sake of clarity rename that, so it won't be confused with the outside stacks?

app/data/Owners.scala
```diff
@@ -0,0 +1,37 @@
...
+    .find(_.hasSSA(SSA(stackName, stageName, appName)))
+    .orElse(all.find(_.hasSSA(SSA(stackName, stageName))))
+    .orElse(all.find(_.hasSSA(SSA(stackName))))
+    .orElse(all.find(_.hasSSA(SSA())))
+}
+
+object Owners extends Owners {
+
+  override def stacks: Set[(String, SSA)] = Set(
+    "phil.wills" -> SSA(), // Default owner of all stacks
```
thomaska: More like an observation (!), intuitively I think we would search more ofter for the owner of an SSA, rather than the list of SSAs for a specific owner, thus storing the information in a mapping SSA -> Owner where Owner could be only the Owner Id? Just a thought!

conf/routes
```diff
@@ -47,3 +47,7 @@ GET        /data/:arn                     controllers.Api.data(arn)
 
 # Map static resources from the /public folder to the /assets URL path
 GET        /assets/*file                  controllers.Assets.at(path="/public", file)
+
+GET        /owners                        controllers.OwnerApi.ownerList
+GET        /owners/for-stack              controllers.OwnerApi.ownerForStack
```
TBonnin: The point of this endpoint is that you don't know the `id` but the would make the following request `/owners/for-stack?stack=discussion&stage=PROD&app=modtools`

app/data/Owners.scala
TBonnin: You are right. `stacks` is not the best name.

app/data/Owners.scala
TBonnin: you are right about what kind of requests would be more often executed. However storage can be done differently. I mostly though about people modifying this file: I found it easier to add/find a team and add/modify the stacks that the other way around, but that's a personal opinion.

sihil: I think that this is the right way around to do it. As @TBonnin says it is possible to reverse this for doing work on it. I notice that there is a unit test to assert that there is only one owner per stack 👍

app/data/Owners.scala
```diff
@@ -0,0 +1,37 @@
...
+  def find(id: String): Option[Owner] = all.find(_.id == id)
+
+  def find(ssa: SSA): Option[Owner] = all.find(_.hasSSA(ssa))
+
+  def forStack(stackName: Option[String], stageName: Option[String], appName: Option[String]): Option[Owner] = all
+    .find(_.hasSSA(SSA(stackName, stageName, appName)))
+    .orElse(all.find(_.hasSSA(SSA(stackName, stageName))))
+    .orElse(all.find(_.hasSSA(SSA(stackName))))
+    .orElse(all.find(_.hasSSA(SSA())))
+}
```
sihil: I think there are some potentially weird behaviours allowed here. For example providing a `stageName` or an `appName` only (without a `stackName`) would return the default owner. I would argue that this is an error and shouldn't return a result at all.

I think the legal queries are:
 - stack only
 - stack and stage
 - stack and app
 - stack, stage and app

Any other combination seems to make little sense and should perhaps return no owner or an error?

conf/routes
sihil: `for-stack` is misleading. Perhaps just `for`?

thomaska:
Thanks a lot for the review and the amendments @sihil !

PR #47, opened 2017-06-20
Add ACM certificates endpoint to Prism

sihil:
Expiring ACM certificates are becoming a problem to it makes sense to have a way of easily identifying certificates that are close to expiring. Adding an endpoint to prism makes this somewhat easier to do at a departmental level.

The new collector needs a couple of new permissions. This has been added to the Prism role CFN template over at https://github.com/guardian/deploy-tools-platform/pull/61. This needs to be merged / rolled out to all of the accounts before this can be deployed.

TODO:
 - [x] Updated prism CFN template rolled out to all accounts

### DIFF
--- a/app/collectors/acmCertificates.scala
+++ b/app/collectors/acmCertificates.scala
@@ -1,0 +1,114 @@
+package collectors
+
+import agent._
+import com.amazonaws.services.certificatemanager.AWSCertificateManagerClientBuilder
+import com.amazonaws.services.certificatemanager.model.{CertificateDetail, DescribeCertificateRequest, ListCertificatesRequest, RenewalSummary, DomainValidation => AwsDomainValidation}
+import controllers.routes
+import org.joda.time.{DateTime, Duration}
+import play.api.mvc.Call
+import utils.Logging
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+object AmazonCertificateCollectorSet extends CollectorSet[AcmCertificate](ResourceType("acm-certificates", Duration.standardMinutes(14L))) {
+  val lookupCollector: PartialFunction[Origin, Collector[AcmCertificate]] = {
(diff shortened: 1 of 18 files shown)


app/collectors/acmCertificates.scala
```diff
@@ -0,0 +1,114 @@
...
+    keyAlgorithm = cert.getKeyAlgorithm,
+    signatureAlgorithm = cert.getSignatureAlgorithm,
+    serial = cert.getSerial,
+    domainValidationOptions = cert.getDomainValidationOptions.asScala.toList.map(DomainValidation.fromApiData),
+    renewalStatus = Option(cert.getRenewalSummary).map(RenewalInfo.fromApiData)
+  )
+}
+
+case class AcmCertificate(
+                              arn: String,
```
adamnfish: I think you need to fix your IntelliJ settings so it indents case class arglists properly. This is an example of where IntelliJ's default is incorrect.

Not super important here, just an FYI.

screen shot 2014-12-23 at 15 48 58

app/collectors/acmCertificates.scala
```diff
@@ -0,0 +1,114 @@
...
+case class RenewalInfo(renewalStatus: String, domainValidationOptions: List[DomainValidation])
+
+object RenewalInfo {
+  def fromApiData(ri: RenewalSummary): RenewalInfo = {
+    RenewalInfo(ri.getRenewalStatus, ri.getDomainValidationOptions.asScala.map(DomainValidation.fromApiData).toList)
+  }
+}
+
+object AcmCertificate {
+  def fromApiData(cert: CertificateDetail, origin: AmazonOrigin) = AcmCertificate(
```
adamnfish: Why's the `origin` argument here?

mchv:
👍

app/collectors/acmCertificates.scala
sihil: It implements the `origin` in `Collector`

PR #48, opened 2017-07-13
Deal with null from getValidationEmails

sihil:
In the real world it seems that `getValidationEmails` can return `null`. This wraps it in `Option` and massages it appropriately.

I've also changed acm-certificates to acmCertificates as it makes jq wrangling easier.


sihil:
@alexduf

PR #49, opened 2017-07-17
add the aws account of each owner, and add more owners

alexduf:
I've also removed some logic since we're relying on hard-coding the values anyway.

The idea is to use the mapping between account -> id to warn owners about imminent certificates expiry.

### DIFF
--- a/app/data/Owners.scala
+++ b/app/data/Owners.scala
@@ -4,13 +4,8 @@ import model._
 
 trait Owners {
   def default: Owner
-  def stacks: Set[(String, SSA)]
 
-  def all: Set[Owner] = stacks
-    .groupBy(_._1)
-    .map { case (ownerId, stacks) =>
-      Owner(ownerId, stacks.map(_._2))
-    }.toSet
+  def all: Set[Owner]
 
   def find(id: String): Option[Owner] = all.find(_.id == id)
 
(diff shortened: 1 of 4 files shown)


test/data/OwnersSpec.scala
```diff
@@ -5,14 +5,6 @@ import org.specs2.mutable._
 
 class OwnersSpec extends Specification {
 
-  "A guardian stack" should {
-    "have only one owner" in {
```
TBonnin: A stack having a single owner is a requirement. We didn't want shared ownership. If I remember correctly that's the main reason the `stacks` were stored as they were.
How is this enforced in this PR?

alexduf: You're right it's not enforced. I figured that because it was hardcoded it would be enough.
I'll add a test that checks the hardcoded values don't contain twice the same ID.

app/model/Owner.scala
```diff
@@ -1,6 +1,7 @@
 package model
 
-case class Owner(id: String, ssas: Set[SSA] = Set.empty) {
+case class Owner(id: String, ssas: Set[SSA] = Set.empty, accounts: Set[String] = Set()) {
   def hasSSA(ssa: SSA): Boolean = ssas.contains(ssa)
+  def email = s"$id@theguardian.com"
```
TBonnin: is it necessary to have this here? Is it used somewhere in prism?

app/data/Owners.scala
```diff
@@ -29,16 +24,22 @@ object Owners extends Owners {
...
+      SSA("cms-fronts"),
+      SSA("elk-new")
+    ), accounts = Set("cmsFronts", "composer", "workflow", "media-service", "cmsSupport")),
+    Owner("dotcom.platform", ssas = Set(SSA("frontend")), accounts = Set("frontend")),
+    Owner("discussiondev", accounts = Set("discussion")),
+    Owner("infosec", accounts = Set("infosec", "security")),
+    Owner("membership.dev", accounts = Set("membership")),
+    Owner("mobile.server.side", accounts = Set("mobile")),
+    Owner("multimedia", accounts = Set("multimedia")),
+    Owner("ophan", accounts = Set("ophan", "tailor"))
```
TBonnin: I assume you have checked with the people concerned when changing some of the owner values (ex: `simon`, `adam`)

TBonnin: Also you seem to have removed `SSA("media-service")`

jennysivapalan:
I agree with @TBonnin I'm little lost between the relationship but I don't actively work on this repo too much which is likely playing a part too. I'll put a note on the line I'm finding a bit hard to understand..

app/data/Owners.scala
```diff
@@ -29,16 +24,22 @@ object Owners extends Owners {
...
-    //TODO: complete list
+  override def all: Set[Owner] = Set(
+    Owner("journalism.dev", ssas = Set(SSA("content-api"), SSA("content-api-preview")), accounts = Set("capi")),
+    Owner("dig.dev.tooling", ssas = Set(SSA("deploy")), accounts = Set("deploy-tools")),
+    Owner("digitalcms.dev", ssas = Set(
+      SSA("flexible"),
+      SSA("workflow"),
+      SSA("cms-fronts"),
+      SSA("elk-new")
+    ), accounts = Set("cmsFronts", "composer", "workflow", "media-service", "cmsSupport")),
```
jennysivapalan: Whats the difference between `ssas` and `accounts` here? Perhaps there is a way to merge the groups together, so `flexible` with `composer`?
I think the `media-service` is missing an `ssa` and `elk-new` is missing an `account`?

alexduf:
Okay I can see it's confusing everyone, let me review my changes than :)

sihil:
Prism already has a strong idea of AWS accounts so I wonder whether owners is the wrong place to put this. It winds up overloading SSAs with accounts. Should these be two separate entities?

What is the ultimate requirement here? What information are we trying to provide?

alexduf:
I am trying to provide a mapping aws account -> owner (email)

Happy to see it anywhere else. In fact should I create a new endpoint for it @sihil ?

PR #53, opened 2017-09-05
Add more detailed ownership for stacks

sihil:
The AMIable weekly e-mails are working well so this fleshes out the ownership list so that it covers more of the known SSA combinations.

I also change the default recipient to be the tooling e-mail list so it's not just @philwills that is bombarded with them.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ scalaVersion in ThisBuild := "2.11.8"
 
 scalacOptions ++= Seq("-unchecked", "-optimise", "-deprecation",
   "-Xcheckinit", "-encoding", "utf8", "-feature", "-Yinline-warnings",
-  "-Xfatal-warnings"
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -8,7 +8,7 @@ scalaVersion in ThisBuild := "2.11.8"
 
 scalacOptions ++= Seq("-unchecked", "-optimise", "-deprecation",
   "-Xcheckinit", "-encoding", "utf8", "-feature", "-Yinline-warnings",
-  "-Xfatal-warnings"
+  "-Xfatal-warnings", "-Ybackend:GenBCode"
```
TBonnin: what is this for?

sihil: See the commit description. In short I hit an optimiser bug that is solved by switching to an alternate (newer) optimiser that is default in 2.12.

PR #55, opened 2017-10-25
Don't store or publish userdata

sihil:
In theory this should be useful for ease of auditing, but in practice there are sometimes credentials in userdata. As such making it so easy to access is a security concern.


philmcmahon:
👍

PR #57, opened 2018-03-13
Remove `tailor`

tomwadeson:
We're removing all traces of `tailor` from our estate and have followed
this thread to: https://prism.gutools.co.uk/owners.


tomwadeson:
Thanks for the +1, @sihil. I don't have perms to merge, however!

PR #70, opened 2019-11-13
WIP - Migrate to play 2.5

mchv:
Play `2.5` is not the latest version but the migration from `2.4` require significant changes on the current application, that I think we should migrate one version after another.

Significant changes includes:

- Filters API change
- No more `Global` object
- `Controllers` can't be object anymore and need to be instantiated.
- `CollectorSet` can't be object anymore and need to be instantiated.
- `Iteratees` has been deprecated in favour or Akka streams.

 
This do not compile yet but wanted to share where I was.
I will continue to update

@akash1810  @philmcmahon @twrichards


philmcmahon:
@jamie-lynch is having a go at a straight jump to Play 2.8 on this - felt like we were too far behind to catch up via incremental migrations

mchv:
@jamie-lynch good luck!

PR #71, opened 2019-12-13
add lambda indexing

twrichards:
For example... 
image
image

... this particular one is filtered on `runtime`, which is a nice touch and will help with Node EOL upgrades.

Original motivation here was actually to ensure `Stage`s used for lambdas were available in RiffRaff - for example in RR we use `DEV` Stage for quite a number of API Gateway backed lambdas (for consistency with play apps where the `DEV` instance runs on our machines - but that's much harder for API Gateway Lambdas).

## MUST add the following permissions to the `PrismRole` in all the AWS accounts

- `lambda:ListFunctions`
- `lambda:ListTags`

... see https://github.com/guardian/deploy-tools-platform/pull/181 (will be StackSet to all the AWD accounts).

### DIFF
--- a/app/collectors/lambda.scala
+++ b/app/collectors/lambda.scala
@@ -1,0 +1,60 @@
+package collectors
+
+import agent._
+import com.amazonaws.services.lambda.model.{FunctionConfiguration, ListFunctionsRequest, ListTagsRequest}
+import com.amazonaws.services.lambda.{AWSLambda, AWSLambdaClientBuilder}
+import controllers.routes
+import play.api.mvc.Call
+import utils.{Logging, PaginatedAWSRequest}
+
+import scala.collection.JavaConversions._
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+object LambdaCollectorSet extends CollectorSet[Lambda](ResourceType("lambda", 1 hour, 5 minutes)) {
+  val lookupCollector: PartialFunction[Origin, Collector[Lambda]] = {
+    case amazon: AmazonOrigin => AWSLambdaCollector(amazon, resource)
+  }
+}
+
+case class AWSLambdaCollector(origin: AmazonOrigin, resource: ResourceType) extends Collector[Lambda] with Logging {
+
+  val client = AWSLambdaClientBuilder.standard()
+    .withCredentials(origin.credentials.provider)
+    .withRegion(origin.awsRegion)
+    .build()
+
+  def crawl: Iterable[Lambda] = {
+    PaginatedAWSRequest.run(client.listFunctions)(new ListFunctionsRequest()).map { lambda =>
+      Lambda.fromApiData(
+        lambda,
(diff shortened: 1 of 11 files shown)


app/controllers/Api.scala
```diff
@@ -169,6 +169,15 @@ trait Api extends Logging {
     singleItem(Prism.instanceAgent, arn)
   }
 
+  // for some bizarre compilation reason this can't be called 'lambdaList'
```
akash1810: If we think `function` isn't intention revealing enough, would `getAllLambdas` work?

app/controllers/Api.scala
```diff
@@ -169,6 +169,15 @@ trait Api extends Logging {
...
+  def functionList = Action.async { implicit request =>
+    itemList(Prism.lambdaAgent, "lambdas")
+  }
+  // for some bizarre compilation reason this can't be called 'lambda'
```
akash1810: If we think `function` isn't intention revealing enough, would `getSingleLambda` work?

app/controllers/Api.scala
twrichards: done https://github.com/guardian/prism/pull/71/commits/1de2c079bd88d8ef0134c6811e832f65d0313af3

app/controllers/Api.scala
twrichards: done https://github.com/guardian/prism/pull/71/commits/1de2c079bd88d8ef0134c6811e832f65d0313af3

app/controllers/Api.scala
mario-galic: retronym (scala contributor) [advises](https://github.com/wix/accord/issues/84#issuecomment-262652711)

> In general, I wouldn't recommend using the experimental GenBCode backend in 2.11.x as there are known bugs which only fixed in 2.12.x.

hence try removing compiler flag

```
-Ybackend:GenBCode
```

twrichards: @mario-galic thank you for your investigation work that got to the bottom of it - one of the weirdest bugs I've ever seen
I've implemented your suggestion which allowed me to make the new function names consistent with all the others - see https://github.com/guardian/prism/pull/71/commits/46fed7f1c2e8968d5ae4e21b132b8dbf638745d9
Thanks

app/collectors/lambda.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+object Lambda {
+
+  def fromApiData(lambda: FunctionConfiguration, client: AWSLambda, region: String, tags: Map[String, String]): Lambda = Lambda(
+    arn = lambda.getFunctionArn,
+    name = lambda.getFunctionName,
+    region,
+    runtime = lambda.getRuntime,
```
twrichards, line 49: @akash1810 @paddyo  I think this is the line that is failing with the Docker lambda

PR #73, opened 2019-12-19
gracefully handle AccessDeniedException for lambda indexing

twrichards:
Follow on from https://github.com/guardian/prism/pull/71

Turns out there are a number of additional AWS accounts which the `PROD` instance of Prism indexes (but not `CODE`) which were not StackSet to include the new permissions `lambda:ListFunctions` and `lambda:ListTags`. Which causes `AccessDeniedException` when trying to index lambdas on those accounts. This PR addresses that by catching the `AccessDeniedException`, logging then returning `Nil`.

### DIFF
--- a/app/collectors/lambda.scala
+++ b/app/collectors/lambda.scala
@@ -1,7 +1,7 @@
 package collectors
 
 import agent._
-import com.amazonaws.services.lambda.model.{FunctionConfiguration, ListFunctionsRequest, ListTagsRequest}
+import com.amazonaws.services.lambda.model.{AWSLambdaException, FunctionConfiguration, ListFunctionsRequest, ListTagsRequest}
 import com.amazonaws.services.lambda.{AWSLambda, AWSLambdaClientBuilder}
 import controllers.routes
 import play.api.mvc.Call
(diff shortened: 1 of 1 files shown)


adamnfish:
Be worth chatting to @sihil about this. My understanding is that it is a deliberate choice to fail the healthcheck if an account can't be scanned. I think the chance that we'd have spotted the permission error if Prism behaved the way this PR works is pretty slim, which is probably why it works this way.

twrichards:
> Be worth chatting to @sihil about this. My understanding is that it is a deliberate choice to fail the healthcheck if an account can't be scanned. I think the chance that we'd have spotted the permission error if Prism behaved the way this PR works is pretty slim, which is probably why it works this way.

@adamnfish I see your point and we will be revisiting this in our discussion about `StackSet`ing some/all of those 15 accounts soon. **However** I think its a shame to further delay this functionality, when it could be providing a useful feature for the majority of our accounts, and this 'graceful' handling only applies to the new lambda indexing. 
At such point in future when all the accounts have the correct permissions (via StackSet or manually) we can revert this PR. _It's perhaps worth pointing out, the result of returning `Nil` is that the accounts without permission do not appear in the list of `sources` in the response._

twrichards:
Also @sihil is off until 20th of January

twrichards:
After some well reasoned discussion IRL with @adamnfish and @katebee we decided its best to press ahead getting the remaining 15 accounts `StackSet'ed (as that has wider benefits) which will avoid the `AccessDeniedException`s actually happening - **so that users of prism can be confident in the comprehensiveness of the results.** 

So although it will delay getting this 'lambda indexing' feature out sooner - I'm closing this PR 😢

adamnfish:
We've chatted IRL and I think the plan is to get assistance from InfoSec to migrate properly to stacksets for the integration, so we can easily update this.

For the record, if this PR was going to go ahead I'd recommend updating the representation so that instead of returning an empty list, it returns something that makes it clear that scanning is not happening. For example, using an `Either[L, List[A]]` where the Left contains something that says this isn't scanned. This would be much safer than silently falling back to an empty list. That would make rendering (and consuming) the response JSON more complex, which I think encourages us to go ahead with making sure all accounts are integrated.

Sorry this is holding the feature up, I agree it's a belter!

PR #81, opened 2020-04-22
Decrease ACM polling inteval to 1 minute

sihil:
Now we have a service that is carrying out validations every minute we should decrease the polling interval to match.


sihil:
Don't think so @philmcmahon - let me know if it balloons.

PR #84, opened 2020-09-28
Migrate to play2.8 & Scala2.13

jamie-lynch:
Attempting to upgrade directly to Play2.8 - as opposed to the strategy employed in #70 - to see if this is less painful that having to do multiple migrations

__Significant changes include__
- Replace `Global` object with `AppComponents` and `AppLoader` (within `PrismApplicationLoader`)
- Changes to the way filters are used
- Controllers have become a class and are instantiated in `PrismComponents` 
- CollectorSet has become a class and is instantiated within `Prism` (the prism controller)
- Akka agents deprecated
- `fileDescriptorLimit` for the systemd service set to 16k (see build.sbt changes)

__Other changes__
- `scala.collection.JavaConversions._` deprecated and replaced with `scala.jdk.CollectionConverters._`
- Guardian management library removed as it appears to be unused (some parts were implemented within the code but the management server was not run and therefore nothing was available)
- `-J-XX:MaxMetaspaceSize` set to 1g from 300m.

__Outstanding__
- [x] ~Currently there's an error in the logs about failing to extract the account number. This is broken for me on master too so is it something that needs fixing (within the scope of this PR)?~ Test this locally with permanent creds from AWS (remove after) to ensure that it still works
- [x] Test the gzip filter is still working
- [x] Fix the tests
- [x] Tidy up `build.sbt`
- [x] Fix the failing build
- [ ] Refactor the Reference and ArnLookup elements to deal better with the need for the prism controller (see comment below) - To make this work we have commented out the JsonInstanceCollector as this doesn't appear to be used in production anywhere. Is this correct?

__Tasks for follow-up PRs__
- Fix `Steam not closed` warnings in `agent/origin.scala` (maybe with `Using(...)`) [Trello Card](https://trello.com/c/DWRLg6cI/292-fix-source-is-not-closed-warnings-in-prism)
- Consider refactoring the `Prism` class to be called `PrismAgents` and not to be within the controllers package [Trello Card](https://trello.com/c/8Yu7WWSK/297-refactor-prismcontroller)
- Consider removing the value of the application secret from version control.

_Reference and ArnLookup refactor_

[...]

PR #87, opened 2020-10-22
Only apply required modifications to Agent.scala and link to source

jamie-lynch:
## What does this change?

Since agents have been deprecated in Akka, we have included the source within the Prism directory as a quick fix rather than migrating to Actors within the [Prism play upgrade](https://github.com/guardian/prism/pull/84). This PR fixes a [comment](https://github.com/guardian/prism/pull/84#discussion_r507694146) at the end of #84 that we should, where possible, leave the Agent code unmodified and include a link to the source.

### DIFF
--- a/app/akka/agent/Agent.scala
+++ b/app/akka/agent/Agent.scala
@@ -1,22 +1,27 @@
-/*
- * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
- */
-
-/*
+/**
  * This file is copied from the Akka project <https://github.com/akka/akka>. It has been included here as Agents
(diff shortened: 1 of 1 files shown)


mbarton:
I'm not sure whether it would be a lot of changes but we have [box](https://github.com/guardian/box), a library which helps with the migration away from Agents

jamie-lynch:
> I'm not sure whether it would be a lot of changes but we have [box](https://github.com/guardian/box), a library which helps with the migration away from Agents

Thanks Michael, I'll definitely take a look at this

philmcmahon (approved):

PR #88, opened 2020-10-26
Crawl all regions

Nirvikalpa108:
## What is the problem? ⭐️

Currently, it is not easy for us to find out if an Amazon resource has been created in a region outside of the Guardian's preferred regions. This is a particular problem for Infosec who get asked questions like,
> "Given an ELB CNAME, can I determine if the resource is in one of our AWS organisation accounts, and if so which account?"

As well as this, we can't rule out if we've been compromised by a hostile third party if we do not have visibility over our resources in all regions. Up until now, Kate sometimes uses Cloudmapper to collect data from AWS accounts but it's not a great experience to run locally: https://github.com/duo-labs/cloudmapper

After speaking to @katebee, we received the following requirements for this work:
1. crawl all regions for AWS assets once a day, or even once per week
2. rule out that there are no AWS assets in other regions
3. once we know which regions we have assets in, Infosec could consider using service control policies to limit guardian users creating assets outside of our desired regions (https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_scps_examples.html#examples_general).

## Describe the solution ⭐️

Therefore, the idea was put forward for prism to crawl all regions, instead of just the mostly eu-west-1 that it crawls currently. This means that it's easy for our Infosec colleagues to find the information they are looking for.

This PR is only the first part of this work. Here, we have:
- split out the shelfLife and refreshPeriod from the ResourceType into a CrawlRate case class, so that we were able to create some central logic that governs all of the crawlRates for the controllers. Previously they were hardcoded into each controller.
- used aws's describeRegions API, which returns all the regions we have access to. See next steps for further work to be done on this.
- we've designed the code that determines crawl rates so that it is easy to test and we've written a test that checks that given a resource type and region, we return the correct crawl rate.
- we've also created layers of overrides, so that prism's config, which lives in dynamoDB can be used to override the logic in the code, both at the Guardian account level and for prism as a whole.

## Next steps ⭐️

Whilst this PR introduces the bulk of the code changes required for this work, there are some significant outstanding issues before this feature is complete:

### AWS rate limits
We're currently hitting rate limits. We tried to solve this by adding a random delay to the scheduled agent, but this only partially worked. My advice for colleagues here would be to:
- understand how the collectors work, so that we know where exactly to add a random delay. It might be the case that where we've added it is not having the most impact. Each collector is initiated when prism starts up and it then starts making its requests. Instead of adding a random delay to the initiation per collector (which is I think what we've done here, although I may be wrong having not looked into it in detail yet), we could look at doing this one level down, at each collector so the timing _their_ requests to all regions are randomised. 

### Access to all regions
The deploy tools account in which prism lives does not have access to all-regions. The describeRegions API returns all the regions we have available to us, which is 16. However with the all-regions option set to true, we should receive 21 regions. The best way to go about this might be to:
- first speak to @katebee about prism getting access to all regions and how this might work
- double check that we can get access to `ap-northeast-3` (Asia Pacific (Osaka-Local)) as when we tried using the aws cli tool with the all-regions flag, that was the one region we didn't have access to that we should have. Here's the aws cli command: `aws ec2 describe-regions --region eu-west-1 --all-regions`.

## How to test  ⭐️
We've tested this PR by deploying it to Prism CODE successfully, with just 2 regions in prism's CODE config (`accounts.aws.regionsDefault`). This is very far away from the 21 regions we hope prism can crawl when this work is complete. A good way to test progress on this work is to slowly increase the number of regions in config, before removing them so that we fall back to the `allRegions` variable introduced in this PR.

## How can we measure success?  ⭐️
Success would be this feature making our Infosec colleagues happy 🥰😍🥰

## Have we considered potential risks?  ⭐️
TBD

### DIFF
--- a/app/controllers/Prism.scala
+++ b/app/controllers/Prism.scala
@@ -4,10 +4,12 @@ import agent.{Accounts, CollectorAgent, SourceStatusAgent}
 import akka.actor.ActorSystem
 import collectors._
 import conf.PrismConfiguration
+import utils.StopWatch
 
 // TODO: Maybe we should refactor this to be PrismAgents and to not be in the controllers package?
 class Prism(prismConfiguration: PrismConfiguration)(actorSystem: ActorSystem) {
-  val sourceStatusAgent = new SourceStatusAgent(actorSystem)
+  val prismRunTimeStopWatch = new StopWatch
+  val sourceStatusAgent = new SourceStatusAgent(actorSystem, prismRunTimeStopWatch)
   val accounts = new Accounts(prismConfiguration)
 
   val lazyStartup: Boolean = prismConfiguration.accounts.lazyStartup
--- a/app/controllers/Prism.scala
+++ b/app/controllers/Prism.scala
@@ -8,7 +8,7 @@ import utils.StopWatch
 
 // TODO: Maybe we should refactor this to be PrismAgents and to not be in the controllers package?
 class Prism(prismConfiguration: PrismConfiguration)(actorSystem: ActorSystem) {
-  val prismRunTimeStopWatch = new StopWatch
+  val prismRunTimeStopWatch = new StopWatch()
   val sourceStatusAgent = new SourceStatusAgent(actorSystem, prismRunTimeStopWatch)
   val accounts = new Accounts(prismConfiguration)
 
--- a/test/ApiSpec.scala
+++ b/test/ApiSpec.scala
@@ -4,7 +4,7 @@ import play.api.libs.json.{JsArray, _}
 import play.api.mvc._
 import play.api.test._
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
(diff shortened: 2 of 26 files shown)


katebee:
> double check that we can get access to ap-northeast-3 (Asia Pacific (Osaka-Local)) as when we tried using the aws cli tool with the all-regions flag, that was the one region we didn't have access to that we should have

This is quite interesting, as it was not clear from the [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions) why this region is not available to most customers. Is appears that only certain AWS-approved customers are able to use the `ap-northeast-3` (Osaka Local) region. According to [this blog post](https://aws.amazon.com/blogs/aws/in-the-works-aws-osaka-local-region-expansion-to-full-region/) the region was intended to serve the requirements for multi-region fault tolerance (without significantly increasing latency) for customers with resources in the Tokyo region:

> Located 400km from the Tokyo Region, the Osaka Local Region has supported customers with applications that require in-country, geographic diversity for disaster recovery purposes that could not be served with the Tokyo Region alone.

The blog post also mentions that AWS are planning to roll it out as a full region in 2021. However, we could consider applying restrictions at the Organisation level to prevent accounts using the `ap-northeast-3` region. Since only a few AWS customers currently have access to this region, it is unlikely we are one of them. Therefore I would suggest that we not try to get Prism to crawl the `ap-northeast-3` region. 😁

This is a really good resource to understand more about AWS global infrastructure: https://infrastructure.aws/

app/controllers/Prism.scala
```diff
@@ -4,10 +4,12 @@ import agent.{Accounts, CollectorAgent, SourceStatusAgent}
 import akka.actor.ActorSystem
 import collectors._
 import conf.PrismConfiguration
+import utils.StopWatch
 
 // TODO: Maybe we should refactor this to be PrismAgents and to not be in the controllers package?
 class Prism(prismConfiguration: PrismConfiguration)(actorSystem: ActorSystem) {
-  val sourceStatusAgent = new SourceStatusAgent(actorSystem)
+  val prismRunTimeStopWatch = new StopWatch
```
philmcmahon, line 11: I think convention is `new Stopwatch()`

test/ApiSpec.scala
```diff
@@ -76,16 +76,20 @@ object ApiSpec extends PlaySpecification with Results {
   }
 
   class TestOrigin extends Origin {
+    private val duration = FiniteDuration(1, "s")
```
philmcmahon, line 79: maybe call this `oneSecond` or something

philmcmahon (approved):
Looks great! 

crawl

PR #90, opened 2020-11-05
Ship logs to central ELK

Nirvikalpa108:
Let's enable prism to log to central ELK! Once this PR is merged, we will test by deploying to CODE and will add screenshots here.

An additional piece of work will be to double check all the log lines in prism and make sure we're happy sending them to central ELK.

Cloudformation changes are here: https://github.com/guardian/deploy-tools-platform/pull/249

PR #91, opened 2020-11-10
adds markers to aid debugging of healthcheck fails

Nirvikalpa108:
## What is the problem?
We're currently working on PR #88, which allows prism to crawl all the AWS regions available to it. This work is facing a problem in that prism is unable to pass the loadbalancer health check when trying to crawl 13 or more regions. This means that we can't deploy this work! 

## What is our solution?
In order to help us better understand this problem we have started shipping prism logs to central ELK (PR #90). In particular we're interested to know how long exactly prism takes to pass its healthcheck. 

This PR introduces Markers so that we can analyse the time taken for prism to crawl certain resources and regions. We're hoping to use the graphing features of Kibana to help us analyse this data. 

Here are the markers that we've introduced:

- resource -> this is the name of the resource eg lambda
- account -> the name of the Guardian account eg deployTools
- region -> eg eu-west-1
- duration -> this is the time it takes for the collector to crawl this resources in milliseconds 
- records -> the number of records crawled

_Note that this work branches off of `crawl-all-regions` and we can discuss actions to take before deploying `crawl-all-regions` to PROD._

## How to test
You could set up local ELK logging or just run prism in DEV mode and check the logs in central ELK.

Here's an example of the logs we generated by running prism in DEV mode:

Screenshot 2020-11-10 at 13 59 33
https://logs.gutools.co.uk/s/devx/goto/6569df9a86249c388f8127eab01e7ea9

Success would be the ability to analyse the fields we've created markers for to understand how long each resource/region crawl is taking.

### DIFF
--- a/app/agent/collector.scala
+++ b/app/agent/collector.scala
@@ -1,17 +1,19 @@
 package agent
 
-import utils.{LifecycleWithoutApp, Logging, ScheduledAgent, StopWatch}
+import utils.{LifecycleWithoutApp, Logging, Marker, ScheduledAgent, StopWatch}
 
 import scala.concurrent.duration._
 import scala.language.postfixOps
 import akka.agent.Agent
 import org.joda.time.DateTime
 import akka.actor.ActorSystem
+import net.logstash.logback.marker.Markers
 
+import scala.jdk.CollectionConverters._
 import scala.concurrent.ExecutionContext
 import scala.util.Random
 
-class CollectorAgent[T<:IndexedItem](val collectorSet: CollectorSet[T], sourceStatusAgent: SourceStatusAgent, lazyStartup:Boolean = true)(actorSystem: ActorSystem) extends CollectorAgentTrait[T] with Logging with LifecycleWithoutApp {
+class CollectorAgent[T<:IndexedItem](val collectorSet: CollectorSet[T], sourceStatusAgent: SourceStatusAgent, lazyStartup:Boolean = true)(actorSystem: ActorSystem) extends CollectorAgentTrait[T] with Logging with Marker with LifecycleWithoutApp {
 
   implicit private val collectorAgent: ExecutionContext = actorSystem.dispatchers.lookup("collectorAgent")
   val collectors: Seq[Collector[T]] = collectorSet.collectors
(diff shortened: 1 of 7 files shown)


app/utils/Marker.scala
```diff
@@ -0,0 +1,5 @@
+package utils
+
+trait Marker {
+  def toMarkerMap: Map[String, Any]
```
Nirvikalpa108, line 4: We've created a trait to avoid repetitive code in the CollectorAgent

philmcmahon (approved):
LGTM!

app/agent/collector.scala
```diff
@@ -32,23 +34,27 @@ class CollectorAgent[T<:IndexedItem](val collectorSet: CollectorSet[T], sourceSt
...
+
   def update(collector: Collector[T], previous:Datum[T]):Datum[T] = {
     val s = new StopWatch
     val datum = Datum[T](collector)
     val timeSpent = s.elapsed
     sourceStatusAgent.update(datum.label)
     datum.label match {
       case l@Label(product, origin, size, _, None) =>
-        log.info(s"Crawl of ${product.name} from $origin successful (${timeSpent}ms): $size records, ${l.bestBefore}")
+        val marker = Markers.appendEntries((origin.toMarkerMap ++ l.toMarkerMap ++ this.toMarkerMap ++ Map("duration" -> timeSpent)).asJava)
```
Nirvikalpa108, line 46: Just to make review easier, I've listed the fields here too:
1. `origin.toMarkerMap` holds our **region** field
2. `l.toMarkerMap` holds our **resource** and **accounts** fields
3. `this.toMarkerMap` is the **records** field on line 37 of this class 
4. and then finally we include the **duration** field here in the `update` function where we have access to the stopwatch

app/agent/origin.scala
```diff
@@ -136,4 +138,6 @@ case class JsonOrigin(vendor:String, account:String, url:String, resources:Set[S
     Json.parse(jsonText)
   }
   val jsonFields = Map("url" -> url)
+
+  override def toMarkerMap: Map[String, Any] = jsonFields
```
Nirvikalpa108, line 142: Where you see `toMarkerMap` set to `jsonFields`, we're doing that just to compile. We don't actually use these values!

PR #92, opened 2020-11-13
Logs time taken between application start and healthcheck passing successfully for each resource

Nirvikalpa108:
## What is the problem?

The root problem we're trying to solve is to have visibility over which AWS resources we have employed across the Guardian estate. Currently we don't have that, which is a security concern. So PR #88 aims to solve this by making Prism crawl all the 16 AWS regions that it has access to. 

In order to deploy this feature, we need Prism to pass its healthcheck, which it is currently only doing with 12 regions. In order to gain more visibility on how long it's taking Prism to perform the initial crawl of all resources in all regions, we have set up log shipping to central ELK (#90) and added helpful markers so that we can analyse the data returned easily (#91).

The logging added thus far is helpful, but it doesn't tell us how long Prism is taking from application start to the healthcheck passing successfully. This PR solves this problem.

## How have we solved it?

We've added logging that indicates once a Prism instance has passed its healthcheck. The logging tells us how long it takes for each resource to be crawled successfully.

## How has this been tested?

We've deployed this to CODE with 12 regions in config.

### These log lines log out once a resource has passed its healthcheck 
Screenshot 2020-11-13 at 10 03 45
https://logs.gutools.co.uk/s/devx/goto/4080c31029e5294b190a50e7eb7c8f0f

Screenshot 2020-11-13 at 10 03 55
https://logs.gutools.co.uk/s/devx/goto/4080c31029e5294b190a50e7eb7c8f0f

### This tells us what % of a resource has been crawled
Screenshot 2020-11-13 at 10 04 22
https://logs.gutools.co.uk/s/devx/goto/283f21942d0ff8c651bf74bb0283acb6

Screenshot 2020-11-13 at 10 04 42
https://logs.gutools.co.uk/s/devx/goto/283f21942d0ff8c651bf74bb0283acb6

### This tells us how long it takes for one resource in one region to be crawled and how many records were crawled
Screenshot 2020-11-13 at 10 04 03
https://logs.gutools.co.uk/s/devx/goto/0367ed51063b611cf71c1b7dc43bed1e

Screenshot 2020-11-13 at 10 04 14
https://logs.gutools.co.uk/s/devx/goto/0367ed51063b611cf71c1b7dc43bed1e

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -6,7 +6,7 @@ import play.api.{ApplicationLoader, BuiltInComponentsFromContext}
 import play.filters.HttpFiltersComponents
 import play.filters.gzip.GzipFilterComponents
 import router.Routes
-import utils.{Lifecycle, Logging, ScheduledAgent, StopWatch}
+import utils.{Lifecycle, Logging, ScheduledAgent}
 
 import scala.collection.mutable
 import scala.concurrent.Future
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -6,7 +6,7 @@ import play.api.{ApplicationLoader, BuiltInComponentsFromContext}
 import play.filters.HttpFiltersComponents
 import play.filters.gzip.GzipFilterComponents
 import router.Routes
-import utils.{Lifecycle, Logging, ScheduledAgent}
+import utils.{Lifecycle, Logging, ScheduledAgent, StopWatch}
 
 import scala.collection.mutable
 import scala.concurrent.Future
--- a/app/agent/collector.scala
+++ b/app/agent/collector.scala
@@ -34,7 +34,7 @@ class CollectorAgent[T<:IndexedItem](val collectorSet: CollectorSet[T], sourceSt
 
   def size: Int = get().map(_.data.size).sum
 
-  override def toMarkerMap: Map[String, Any] = Map("records" -> size)
+  override def toMarkerMap: Map[String, Any] = Map("records" -> size, "durationType" -> "crawl")
 
   def update(collector: Collector[T], previous:Datum[T]):Datum[T] = {
     val s = new StopWatch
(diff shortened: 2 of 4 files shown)


app/AppComponents.scala
```diff
@@ -6,7 +6,7 @@ import play.api.{ApplicationLoader, BuiltInComponentsFromContext}
 import play.filters.HttpFiltersComponents
 import play.filters.gzip.GzipFilterComponents
 import router.Routes
-import utils.{Lifecycle, Logging, ScheduledAgent}
+import utils.{Lifecycle, Logging, ScheduledAgent, StopWatch}
```
philmcmahon, line 9: probably unnecessary import here

app/agent/collector.scala
```diff
@@ -43,7 +46,7 @@ class CollectorAgent[T<:IndexedItem](val collectorSet: CollectorSet[T], sourceSt
     sourceStatusAgent.update(datum.label)
     datum.label match {
       case l@Label(product, origin, size, _, None) =>
-        val marker = Markers.appendEntries((origin.toMarkerMap ++ l.toMarkerMap ++ this.toMarkerMap ++ Map("duration" -> timeSpent)).asJava)
+        val marker = Markers.appendEntries((origin.toMarkerMap ++ l.toMarkerMap ++ this.toMarkerMap ++ Map("CrawlDuration" -> timeSpent)).asJava)
```
philmcmahon, line 49: we've got CrawlDuration here and `duration` in the other place - maybe worth making these field names super explicity

philmcmahon (approved):

Nirvikalpa108:
We're already learning some really interesting information, for example the `buckets` resource may be taking the longest to crawl, despite the region being crawled:

Screenshot 2020-11-13 at 12 23 06
https://logs.gutools.co.uk/s/devx/goto/afecf8e33db8d22ea5543787cda087ea

app/agent/collector.scala
Nirvikalpa108: Great, thanks Phil! 👍🏽😀

I've added a `durationType` field so that we can keep using the `duration` field where required and add a unique type to make the value of that marker explicit!

PR #93, opened 2020-11-17
Revert "Crawl all regions"

jamie-lynch:
This seems to have made Prism unstable; there were a few build failures in RiffRaff. We've redeployed the last build in the mean time. 

Also noticed a couple of differences between CODE and PROD: 
* We've added the LoggingSteamName to the Dynamo config in PROD
* Looking at the parameters, CODE is using t3.large whilst PROD is still on t3.small.

@philmcmahon @Nirvikalpa108


philmcmahon:
👍  merging this for now whilst we investigate

PR #95, opened 2020-11-19
lambda moved onto aws v2

Nirvikalpa108:
### What does this change?
We've moved the Lambda Collector to v2 of the AWS SDK. We've done this to aid prism in crawling all regions (PR #88). Our hypothesis is that if we can move prism onto v2 of the SDK, eventually moving to use of the asynchronous clients, we can speed up Prism's time to pass its healthcheck.

### How to test
We tested this by deploying to CODE and checking that prism is crawling lambdas and completing its lambda healthcheck successfully.

### DIFF
--- a/app/AppLoader.scala
+++ b/app/AppLoader.scala
@@ -1,8 +1,7 @@
 import com.amazonaws.regions.Regions
 import com.amazonaws.util.EC2MetadataUtils
-import conf.{AWS, DynamoConfiguration, FileConfiguration, Identity, LogConfiguration}
-import play.api.Configuration
-import play.api._
+import conf._
+import play.api.{Configuration, _}
 import utils.{AWSCredentialProviders, Logging}
 
 import scala.util.Try
@@ -48,6 +47,12 @@ class AppLoader extends ApplicationLoader with Logging {
 
     val contextWithConfig = context.copy(initialConfiguration = combinedConfig)
 
-    new AppComponents(contextWithConfig).application
+    try {
+      new AppComponents(contextWithConfig).application
+    } catch {
+      case t: Throwable =>
+        log.error("Unable to initialise prism", t)
+        throw t
+    }
   }
 }
--- a/app/agent/origin.scala
+++ b/app/agent/origin.scala
@@ -5,13 +5,18 @@ import java.net.{URI, URL, URLConnection, URLStreamHandler}
 
 import collectors.Instance
 import com.amazonaws.auth._
-import com.amazonaws.auth.profile.ProfileCredentialsProvider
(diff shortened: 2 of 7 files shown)


app/AppLoader.scala
```diff
@@ -48,6 +47,12 @@ class AppLoader extends ApplicationLoader with Logging {
 
     val contextWithConfig = context.copy(initialConfiguration = combinedConfig)
 
-    new AppComponents(contextWithConfig).application
+    try {
+      new AppComponents(contextWithConfig).application
+    } catch {
+      case t: Throwable =>
+        log.error("Unable to initialise prism", t)
+        throw t
```
Nirvikalpa108, line 55: Whilst debugging this work we had to go onto the instance and run `$ journalctl -u prism` to get the error that helped us understand why prism was not starting. In order for this to not happen again, we're catching such throwables and logging them out to central ELK for easier debugging.

build.sbt
```diff
@@ -39,6 +40,10 @@ lazy val root = (project in file("."))
     buildInfoPackage := "prism",
     libraryDependencies ++= Seq(
       "com.google.code.findbugs" % "jsr305" % "2.0.0",
+      "software.amazon.awssdk" % "lambda" % awsVersionTwo,
+      "software.amazon.awssdk" % "auth" % awsVersionTwo,
+      "software.amazon.awssdk" % "sts" % awsVersionTwo,
+      "com.beust" % "jcommander" % "1.75", // deal with security vulnerability introduced by aws sdk
```
Nirvikalpa108, line 46: Ideally, let's delete this dependency once the vulnerability in aws sdk v2 is fixed! I'm not sure how we can remind ourselves to do this? Perhaps by adding a TODO?

app/agent/origin.scala
```diff
@@ -56,20 +61,57 @@ trait Origin extends Marker {
 }
 
 case class Credentials(accessKey: Option[String], role: Option[String], profile: Option[String], region: String)(secretKey: Option[String]) {
```
Nirvikalpa108, line 63: This case class has ballooned because we're currently running both v1 and v2 of the SDK at the same time. This will look much better once we move all the collectors onto v2!😅

build.sbt
philmcmahon: yep TODO in the comments sounds good, maybe with a link to some post discussing it

philmcmahon (approved):
LGTM 👍  only thing I think it would be helpful to include a link to [snyk](https://snyk.io/vuln/maven:com.beust%3Ajcommander) or wherever you found out about the vulnerability

PR #96, opened 2020-11-20
Adds instance and build ids to the logs

Nirvikalpa108:
Here we add the instance id and build id to prism logs, to aid developer debugging on the project.


Nirvikalpa108:
Here are the instance id and build id in the logs after deploying to CODE ☺️
Screenshot 2020-11-20 at 11 53 39

philmcmahon (approved):

sihil (approved):

PR #97, opened 2020-11-24
Bucket Collector to AWS SDK v2

Nirvikalpa108:
### What does this change?

### TODO
1. After deploy, check the logs to see if we're crawling buckets at all. If not, then debug what `Region.toString` vs `GetBucketLocationResponse.locationConstraintAsString()` is in `Bucket.scala:53`, as this could be the culprit.

1. Redo teamcity build after committing change within github ui

### DIFF
--- a/app/collectors/bucket.scala
+++ b/app/collectors/bucket.scala
@@ -1,8 +1,10 @@
 package collectors
 
 import agent._
-import com.amazonaws.services.s3.{AmazonS3, AmazonS3ClientBuilder}
-import com.amazonaws.services.s3.model.{AmazonS3Exception, ListBucketsRequest, Bucket => AWSBucket}
+import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.s3.model.{GetBucketLocationRequest, ListBucketsRequest, S3Exception, Bucket => AWSBucket}
+
+import scala.language.reflectiveCalls
 import conf.AWS
 import controllers.routes
 import org.joda.time.DateTime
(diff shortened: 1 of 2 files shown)


app/collectors/bucket.scala
```diff
@@ -43,23 +46,23 @@ object Bucket {
...
 
-  def fromApiData(bucket: AWSBucket, client: AmazonS3): Option[Bucket] = {
-    val bucketName = bucket.getName
+  def fromApiData(bucket: AWSBucket, client: S3Client, originRegion: Region): Option[Bucket] = {
+    val bucketName = bucket.name
     try {
-      val bucketRegion = client.getBucketLocation(bucket.getName)
-      if (bucketRegion == client.getRegionName) {
+      val bucketRegion = client.getBucketLocation(GetBucketLocationRequest.builder().bucket(bucketName).build()).locationConstraintAsString()
+      if (bucketRegion == originRegion.toString) {
```
sihil, line 53: I'd prefer using `id` here. As it happens `toString` is the same and will work, but `id` would be cleaner and more explicit.
```suggestion
      if (bucketRegion == originRegion.id) {
```

sihil (changes requested):
A minor improvement otherwise looks good.

sihil (approved):

Nirvikalpa108:
After switching to java instance time from joda datetime and running `$ diff -I href <(jq -S '.data.buckets | sort_by(.arn)' pBuckets.json) <(jq -S '.data.buckets | sort_by(.arn)' cBuckets.json)`, the only difference is that CODE does not include milliseconds. We don't think any application that ingests prism data is likely to be checking miliseconds, so this should be good to deploy. 

```
<     "createdTime": "2020-06-25T07:06:41.000Z",
---
>     "createdTime": "2020-06-25T07:06:41Z",
```

PR #98, opened 2020-11-24
Add /sources/accounts endpoint.

philmcmahon:
## What does this change?
Introduces a new endpoint to list aws accounts containing instances indexed by prism. The output looks like this: 
Screenshot 2020-11-24 at 15 59 44

## How to test

This is currently on CODE https://prism.code.dev-gutools.co.uk/sources/accounts

We can use this endpoint to e.g. power UI elements that need a list of AWS accounts. It should also be useful for quickly answering the question "what account is 12312312"?

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -200,6 +200,19 @@ class Api (cc: ControllerComponents, prismDataStore: Prism, prismConfiguration: 
     def mainclassList = summary[Instance](prismDataStore.instanceAgent, i => i.mainclasses.map(Json.toJson(_)), "mainclasses")
     def stackList = summaryFromTwo[Instance, Lambda](prismDataStore.instanceAgent, stackExtractor, prismDataStore.lambdaAgent, stackExtractor, "stacks")(prismConfiguration.stages.ordering)
     def stageList = summaryFromTwo[Instance, Lambda](prismDataStore.instanceAgent, stageExtractor, prismDataStore.lambdaAgent, stageExtractor, "stages")(prismConfiguration.stages.ordering)
+    def instanceAccounts: Action[AnyContent] = Action.async { implicit request =>
+      ApiResult.noSource {
+        val accounts = prismDataStore.instanceAgent.get().map{c =>
+          val accountName = c.label.origin.account
+          val accountNumber = c.label.origin match {
+            case a: AmazonOrigin => a.accountNumber
+            case _ => None
+          }
+          AWSAccount(accountNumber, accountName)
+        }.toList
+        Json.toJson(accounts.distinct)
+      }
+    }
     def regionList = summary[Instance](prismDataStore.instanceAgent, i => Some(Json.toJson(i.region)), "regions")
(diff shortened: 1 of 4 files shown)


sihil:
I'm not sure we should constrain ourselves to instances here. I think it would be better derived from sources and that it might better live at `/sources/accounts` for that reason.

katebee:
> We can use this endpoint to e.g. power UI elements that need a list of AWS accounts. It should also be useful for quickly answering the question "what account is 12312312"?

I ask myself this frequently, often followed by "what is the email group for the team that manages this account?" 😁 
We currently don't have a great way of mapping account names to account IDs. While this would be helpful, is it worth having a discussion around what Prism collects, what we consider private information, and what we might want to skip collecting?

philmcmahon:
Thanks @sihil - I've updated this now.

@katebee In terms of reducing the amount of private info in prism, it might be good to review how much of the infor we need. I think my preference would be to improve the security of prism. I could imagine using prism in future to find out e.g. 'what account is X param store property in'

philmcmahon:
For this specific endpoint, it doesn't expose info that isn't already exposed on /sources - it just results in a slightly smaller response (2kb instead of 280).

app/controllers/Api.scala
```diff
@@ -123,6 +123,20 @@ class Api (cc: ControllerComponents, prismDataStore: Prism, prismConfiguration:
       Api.singleItem(prismDataStore.instanceAgent, arn)
     }
 
+    def instanceAccounts: Action[AnyContent] = Action.async { implicit request =>
+      ApiResult.noSource {
+        val accounts = prismDataStore.instanceAgent.get().map { c =>
```
sihil: I don't think this should hang off the `instanceAgent`. There could be an account that we don't crawl instances for.

sihil: See sources on line 74.

philmcmahon: I think the diff might not have updated - this now uses the sources agent

sihil: So it does. Weird.

app/controllers/Api.scala
```diff
@@ -81,6 +81,20 @@ class Api (cc: ControllerComponents, prismDataStore: Prism, prismConfiguration:
       }
     }
 
+    def sourceAccounts: Action[AnyContent] = Action.async { implicit request =>
+      ApiResult.noSource {
```
sihil, line 85: You could use `filter` here instead of noSource. The left hand side of the returned Map is the label of the source - see the `sources` endpoint above. That would make it filterable and incorporate the staleness checks.
Having said that, there probably isn't a reason to filter and will never be stale as it comes purely from config so 🤷‍♂️

sihil (approved):
Looks good unless you want to make it possible to filter.

app/controllers/Api.scala
philmcmahon: In the end I decided against this as I only want one result per account - rather than one result per account per source. I think this makes sense, anyone wanting full filtering can use the /sources endpoint

PR #101, opened 2020-11-25
Images Collector to AWS SDK v2

Nirvikalpa108:
### What does this change?

## TODO
remove `account.amis` block from CODE config after testing

### DIFF
--- a/app/collectors/image.scala
+++ b/app/collectors/image.scala
@@ -1,16 +1,15 @@
 package collectors
 
 import agent._
-import com.amazonaws.services.ec2.{AmazonEC2, AmazonEC2ClientBuilder}
-import com.amazonaws.services.ec2.model.{DescribeImagesRequest, Filter, Image => AWSImage}
 import conf.AWS
 import controllers.routes
 import org.joda.time.DateTime
 import play.api.mvc.Call
+import software.amazon.awssdk.services.ec2.Ec2Client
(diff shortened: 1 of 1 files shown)


app/collectors/image.scala
```diff
@@ -22,43 +21,47 @@ class ImageCollectorSet(accounts:Accounts) extends CollectorSet[Image](ResourceT
...
     .build()
 
   def crawl: Iterable[Image] = {
-    val result = client.describeImages(new DescribeImagesRequest()
-      .withFilters(
-        new Filter("owner-id", Seq(origin.accountNumber.get).asJava),
-        new Filter("image-type", Seq("machine").asJava)
+    val ownerIdFilter = Filter.builder().name("owner-id").values(origin.accountNumber.get).build()
+    val imageTypeFilter = Filter.builder().name("image-type").values("machine").build()
+    //TODO
```
sihil, line 34: 🔍

sihil (approved):
Apart from the errant TODO, LGTM

Nirvikalpa108:
We tested this by changing CODE config to match PROD and checking the difference in output between the two running `$ diff -I href <(jq -S '.data.images | sort_by(.arn)' pImages.json) <(jq -S '.data.images | sort_by(.arn)' cImages.json)`. No diff, which is great! 🎉

We also checked the lines (although the output is on one line) so ran `$ cat cImages.json | jq | wc -l` to confirm that both CODE and PROD output had the same count. 😀

PR #103, opened 2020-11-26
Server Certificate Collector to AWS SDK v2

Nirvikalpa108:
## What does this change?

### DIFF
--- a/app/collectors/serverCertificates.scala
+++ b/app/collectors/serverCertificates.scala
@@ -1,20 +1,20 @@
 package collectors
 
+import java.time.Instant
+
 import agent._
-import com.amazonaws.services.identitymanagement.{AmazonIdentityManagement, AmazonIdentityManagementClientBuilder}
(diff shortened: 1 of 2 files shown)


sihil (dismissed):
I don't think we use server certificates any more. That said there are still dozens left lying around so let's keep this. LGTM.

Nirvikalpa108:
We failed to deploy this to CODE and received the following error from the logs in central ELK:
```
software.amazon.awssdk.core.exception.SdkClientException: 
Received an UnknownHostException when attempting to interact with a service. 
See cause for the exact endpoint that is failing to resolve. 
This specific service is global in the same partition as the region configured on this client (us-east-1). 
If this is the first time you're trying to talk to this service in this region, 
you should try configuring the global region on your client, instead: aws-global
```
Deploying the aws sdk upgrade to the server certificate collector is likely to be a larger piece of work, because prism will need to know about the `aws-global` region and the collectors that take this.

PR #105, opened 2020-11-27
Route 53 Collector to AWS SDK v2

Nirvikalpa108:
## What does this change?


sihil (approved):
LGTM :shipit:

sihil:
Route53 is also a global service so perhaps needs to be solved in the same way as server certs.

Nirvikalpa108:
We're receiving the same error as PR #103: 
```
software.amazon.awssdk.core.exception.SdkClientException: 
Received an UnknownHostException when attempting to interact with a service. 
See cause for the exact endpoint that is failing to resolve. 
This specific service is global in the same partition as the region configured on this client (us-east-1). 
If this is the first time you're trying to talk to this service in this region, 
you should try configuring the global region on your client, instead: aws-global
```
We'll need to do some more work to resolve this.

PR #107, opened 2020-11-27
Reservation Collector to AWS SDK v2

Nirvikalpa108:
### What does this change?

### DIFF
--- a/app/collectors/reservation.scala
+++ b/app/collectors/reservation.scala
@@ -3,17 +3,15 @@ package collectors
 import java.time.Instant
 
 import agent._
-import software.amazon.awssdk.services.ec2.Ec2Client
-import software.amazon.awssdk.services.ec2.model.{DescribeReservedInstancesRequest, ReservedInstances, RecurringCharge => AwsRecurringCharge}
 import conf.AWS
 import controllers.routes
-import org.joda.time.DateTime
 import play.api.mvc.Call
+import software.amazon.awssdk.services.ec2.Ec2Client
+import software.amazon.awssdk.services.ec2.model.{DescribeReservedInstancesRequest, ReservedInstances, RecurringCharge => AwsRecurringCharge}
 import utils.Logging
 
 import scala.jdk.CollectionConverters._
 import scala.language.postfixOps
-import scala.util.Try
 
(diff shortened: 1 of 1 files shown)


app/collectors/reservation.scala
```diff
@@ -50,35 +52,35 @@ case class Reservation(
...
-      offeringType = reservationInstance.getOfferingType,
-      startTime = Try(new DateTime(reservationInstance.getStart)).toOption,
-      endTime = Try(new DateTime(reservationInstance.getEnd)).toOption
+      state = reservationInstance.stateAsString,
+      currencyCode = reservationInstance.currencyCodeAsString,
+      duration = reservationInstance.duration,
+      instanceTenancy = reservationInstance.instanceTenancyAsString,
+      offeringType = reservationInstance.offeringTypeAsString,
+      startTime = Try(reservationInstance.start).toOption,
+      endTime = Try(reservationInstance.end).toOption
```
sihil, line 83: In the previous implementation the `Try` was needed as `new DateTime` can throw an exception (if the value being passed in isn't a valid DateTime. With the new SDK there is no risk of an exception so we can remove the `Try`. If `start` and `end` can be null then we might need to wrap it with `Option(<value>)` but otherwise we could just pass it through directly and remove the `Option` type from `Reservation`.

sihil: Look at the number of instances of a field that isn't an option (always there) and our time fields which are optional (will be absent if the date format is bad or the value is null):
```
❯ curl -s https://prism.gutools.co.uk/reservations | jq | grep offeringType | wc -l
     928
❯ curl -s https://prism.gutools.co.uk/reservations | jq | grep startTime | wc -l
     928
❯ curl -s https://prism.gutools.co.uk/reservations | jq | grep endTime | wc -l
     928
```

From this I reckon `startTime` and `endTime` can't be null so it's safe to remove the option entirely.

sihil (changes requested):
Let's remove the optionality of the times entirely. Otherwise this is ready.

app/collectors/reservation.scala
Nirvikalpa108: This explanation is amazing! Thank you! 💯👏🏽🥇👍🏽😀

PR #110, opened 2020-12-03
Adds global vs regional concept to collectors

Nirvikalpa108:
## What does this change?
The AWS SDK 2 introduces the idea of a global region for certain services. For example, Route 53 (Amazon's DNS service). Here, we add this concept to Prism's collectors which can now be Global or Regional. This change means that we can complete the migration process to sdk 2. 

No more error messages when we deploy PRs #105 and #103.

### DIFF
--- a/app/agent/model.scala
+++ b/app/agent/model.scala
@@ -33,8 +33,17 @@ sealed trait AwsRegionType
 case object Global extends AwsRegionType
 case object Regional extends AwsRegionType
 
-/** The awsRegionType specifies whether an AWS collector is global or regional. The AWS_GLOBAL region is required for
- * global services, such as Route 53. */
+/** A CollectorSet knows how to create a set of collectors for a given resource type that typically
+ *  spans multiple accounts, which can be of different underlying platforms. 
+ *  A CollectorSet creates an appropriate set of Collector instances for each account and region.
+ *
+ * @param resourceType the name of the resource that this CollectorSet is responsible for
+ * @param accounts the set of accounts to collect this resource from
+ * @param awsRegionType some resourceTypes in AWS have a single Global instance instead of Regional 
+ * instances. If a CollectorSet processes `AmazonOrigin` origins then you should specify whether the AWS 
+ * collector is global (such as Route53) or regional (such as EC2 instances).
+ * @tparam T the class that represents a collected instance of the resource
+ */
 abstract class CollectorSet[T](val resource:ResourceType, accounts: Accounts, val awsRegionType: Option[AwsRegionType]) extends Logging {
   /** Create a collector for the given origin (this is a partial function because not all collectors support
    *  all types of origin */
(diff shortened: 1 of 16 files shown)


app/agent/model.scala
```diff
@@ -28,12 +29,30 @@ trait IndexedItemWithStack extends IndexedItem {
 
 case class AWSAccount(accountNumber: Option[String], accountName: String)
 
+sealed trait AwsRegionType
+case object Global extends AwsRegionType
+case object Regional extends AwsRegionType
+
 abstract class CollectorSet[T](val resource:ResourceType, accounts: Accounts) extends Logging {
```
sihil, line 36: Another option that might reduce boilerplate:
```suggestion
abstract class CollectorSet[T](val resource:ResourceType, accounts: Accounts, val awsRegionType: Option[AwsRegionType]) extends Logging {
```

Just another way of expressing the same thing, but would reduce repetition as the regionType would be passed in as a parameter. The scaladoc could then be moved to describe the CollectorSet.

sihil (commented):
🎉 Looks great.

I had a couple of minor thoughts whilst reviewing - one was to move `awsRegionType` from being field on the class to being a param which will reduce boiler plate in each class. Have a look and see what you think.

app/agent/model.scala
```diff
@@ -28,12 +29,30 @@ trait IndexedItemWithStack extends IndexedItem {
 
 case class AWSAccount(accountNumber: Option[String], accountName: String)
 
-abstract class CollectorSet[T](val resource:ResourceType, accounts: Accounts) extends Logging {
+sealed trait AwsRegionType
+case object Global extends AwsRegionType
+case object Regional extends AwsRegionType
+
+/** The awsRegionType specifies whether an AWS collector is global or regional. The AWS_GLOBAL region is required for
+ * global services, such as Route 53. */
```
sihil, line 37:
```suggestion
/** A CollectorSet knows how to create a set of collectors for a given resource type that typically
 *  spans multiple accounts, which can be of different underlying platforms. 
 *  A CollectorSet creates an appropriate set of Collector instances for each account and region.
 *
 * @param resourceType the name of the resource that this CollectorSet is responsible for
 * @param accounts the set of accounts to collect this resource from
 * @param awsRegionType some resourceTypes in AWS have a single Global instance instead of Regional 
 * instances. If a CollectorSet processes `AmazonOrigin` origins then you should specify whether the AWS 
 * collector is global (such as Route53) or regional (such as EC2 instances).
 * @tparam T the class that represents a collected instance of the resource
 */
```

sihil (approved):
I think this looks good and ready to ship. I've made an improvement to the scaladoc on the class to better document the individual parameters and type parameter.

sihil (approved):

sihil (approved):

Nirvikalpa108:
Tested on CODE with `$ diff -I href <(jq -S '.data | .["server-certificates"] | sort_by(.arn) | unique_by(.arn)' pServerCert.json) <(jq -S '.data | .["server-certificates"] | sort_by(.arn) | unique_by(.arn)' cServerCert.json)` 🎉

sihil:
Worth noting that this work revealed a bug that was actually crawlling route53 and server-certificates twice so we had duplicates in Prism. 🎉

PR #113, opened 2020-12-15
makes s3 collector global

Nirvikalpa108:
## The problem
- we want Prism to crawl more regions (other than the current eu-west-1 and us-east-1) so that we have more visibility over our aws infrastructure. This is important for security reasons. (See more here: PR #88)
- it's hard for Prism to crawl all the aws regions (it's a lot of requests to make to AWS, which takes a long time, which means that prism is failing to pass its healthcheck), so we decided to minimise the number of regions for prism to crawl. We're doing this by shutting down regions that Guardian accounts are able to create new services in (eg https://github.com/guardian/service-control-policies/pull/3) 
-  **The problem is that we have lots of S3 buckets across many different regions** (see the results of a crawl done in November of [all aws regions](https://docs.google.com/spreadsheets/d/1xn9Yr-3_1yWG2qvsHrdXviy6oevGO7gJVt0WCFFQWCg/edit?usp=sharing)). This is preventing us from shutting down access to these regions. 

## The solution
The good news is that AWS's S3 service is special ✨ in that S3 is a kind of pseudo global service. You'll notice when you go into the aws console, that you are unable to select a region for S3, that's because AWS offer resources in S3 globally. 

**Therefore, we don't need to crawl each region to access all of our S3 buckets, we only need to crawl one.** This is fantastic news for us, because it means that we can get access to all our S3 buckets by just crawling `eu-west-1`.💪🏽👏🏽🥳

The advantage of this, to reiterate, is that we can shut down even more regions using our beloved service control policies (https://github.com/guardian/service-control-policies 💝), which means that prism has even fewer regions to crawl.

## How we've tested this
We tested this by deploying to CODE and running `$ curl -s  https://prism.code.dev-gutools.co.uk/buckets | jq .data.buckets[].region | sort | uniq -c | sort -nr` which gives:

```
1114 "eu-west-1"
 117 "EU"
  74 "us-east-1"
  18 "ap-southeast-2"
  12 "us-west-2"
  11 "us-west-1"
   8 "eu-west-2"
   4 "us-east-2"
   4 "eu-central-1"
   1 "eu-west-3"
```

Running the same bash command on current PROD results in the following result `1114 "eu-west-1"`. 

Not only does this change mean that we are able to get information on S3 buckets in all regions by crawling just one, it also means that we have correcting the functionality of our S3 collector, which was previously only returning buckets in `eu-west-1`.💪🏽

## We fixed the createdTime field too - but we had to crawl one extra region
We noticed that we were receiving incorrect creation dates for the S3 buckets, which was introduced prior to this PR and is a known S3 bug. 

To mitigate this, we created another client whose region is set to `us-east-1`, so that we can receive the correct `creationDate` as per this [Stackoverflow post](https://stackoverflow.com/questions/54353373/getting-incorrect-creation-dates-using-aws-s3). 

We tested this change after deploying to CODE with the following bash commands:

_In PROD the createdTime is broken with most of the buckets shown as being created in 2020, which is wrong. This is caused by the region of the S3 Client being set to anything other than `us-east-1`_ 😔
```
$ curl -s https://prism.gutools.co.uk/buckets | jq .data.buckets[].createdTime | sort | cut -d- -f1 | uniq -c
 101 "2019
1013 "2020
```

_In CODE we see a full range of created time dates of the buckets_ 🎉 😃 🎉
```
$ curl -s https://prism.code.dev-gutools.co.uk/buckets | jq .data.buckets[].createdTime | sort | cut -d- -f1 | uniq -c
   2 "2009
  34 "2012
  36 "2013
  86 "2014
 167 "2015
 185 "2016
 132 "2017
 216 "2018
 342 "2019
 163 "2020
```

### DIFF
--- a/app/collectors/bucket.scala
+++ b/app/collectors/bucket.scala
@@ -48,11 +48,11 @@ object Bucket {
   def fromApiData(bucket: AWSBucket, client: S3Client): Option[Bucket] = {
     val bucketName = bucket.name
     try {
-      val bucketRegion = client.getBucketLocation(GetBucketLocationRequest.builder.bucket(bucketName).build).locationConstraintAsString
+      val bucketRegion = Option(client.getBucketLocation(GetBucketLocationRequest.builder.bucket(bucketName).build).locationConstraintAsString)
       Some(Bucket(
         arn = arn(bucketName),
         name = bucketName,
-        region = bucketRegion,
+        region = bucketRegion.getOrElse(Region.US_EAST_1.id),
         createdTime = Try(bucket.creationDate).toOption
       ))
     } catch {
--- a/app/collectors/bucket.scala
+++ b/app/collectors/bucket.scala
@@ -28,7 +28,7 @@ case class AWSBucketCollector(origin: AmazonOrigin, resource: ResourceType, craw
   // https://stackoverflow.com/questions/46769493/how-enable-force-global-bucket-access-in-aws-s3-sdk-java-2-0
   val s3Configuration = S3Configuration.builder.useArnRegionEnabled(true).build
 
-  // The region of the S3 Client is hardcoded to EU-WEST-1, because AWS-Global does not work with S3.
+  // The region of the S3 Client is hardcoded to EU-WEST-1, because AWS-Global and US-EAST-1 do not return all buckets.
   // We decided to hardcode this here, instead of creating another enum for simplicity 
   val client = S3Client
     .builder
@@ -38,8 +38,8 @@ case class AWSBucketCollector(origin: AmazonOrigin, resource: ResourceType, craw
     .serviceConfiguration(s3Configuration)
     .build
 
-  // This second S3 client, with a region of US-EAST-1, gives us the correct createdTime value as documented here:
-  // https://stackoverflow.com/questions/54353373/getting-incorrect-creation-dates-using-aws-s3
+  // This second S3 client, with a region of US-EAST-1, gives us the correct createdTime value unlike the other regions,
+  // as documented here: https://stackoverflow.com/questions/54353373/getting-incorrect-creation-dates-using-aws-s3
   val clientForCorrectCreatedTime = S3Client
     .builder
     .credentialsProvider(origin.credentials.provider)
--- a/app/collectors/bucket.scala
+++ b/app/collectors/bucket.scala
@@ -30,7 +30,7 @@ case class AWSBucketCollector(origin: AmazonOrigin, resource: ResourceType, craw
   val client = S3Client
     .builder
     .credentialsProvider(origin.credentials.provider)
-    .region(Region.US_EAST_1)
+    .region(Region.EU_WEST_1)
     .overrideConfiguration(AWS.clientConfig)
     .serviceConfiguration(s3Configuration)
     .build
--- a/app/collectors/bucket.scala
+++ b/app/collectors/bucket.scala
@@ -7,7 +7,7 @@ import conf.AWS
 import controllers.routes
 import play.api.mvc.Call
 import software.amazon.awssdk.regions.Region
-import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.s3.{S3Client, S3Configuration}
 import software.amazon.awssdk.services.s3.model.{GetBucketLocationRequest, ListBucketsRequest, S3Exception, Bucket => AWSBucket}
 import utils.Logging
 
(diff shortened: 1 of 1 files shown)


app/collectors/bucket.scala
```diff
@@ -7,72 +7,94 @@ import conf.AWS
...
 import utils.Logging
 
 import scala.jdk.CollectionConverters._
 import scala.language.{postfixOps, reflectiveCalls}
-import scala.util.Try
 import scala.util.control.NonFatal
 
 
-class BucketCollectorSet(accounts: Accounts) extends CollectorSet[Bucket](ResourceType("bucket"), accounts, Some(Regional)) {
+class BucketCollectorSet(accounts: Accounts) extends CollectorSet[Bucket](ResourceType("bucket"), accounts, Some(Global)) {
```
Nirvikalpa108, line 19: We had to create a new enum value for all of Prism's collectors when we moved to AWS SDK 2 (https://github.com/guardian/prism/pull/110). We've made the S3 Bucket Collector a Global service, so that we can crawl fewer regions (the Regional collectors crawl every region).

app/collectors/bucket.scala
```diff
@@ -7,72 +7,96 @@ import conf.AWS
...
     .credentialsProvider(origin.credentials.provider)
-    .region(origin.awsRegionV2)
+    .region(Region.EU_WEST_1)
     .overrideConfiguration(AWS.clientConfig)
+    .serviceConfiguration(s3Configuration)
+    .build
+
+  // This second S3 client, with a region of US-EAST-1, gives us the correct createdTime value as documented here:
+  // https://stackoverflow.com/questions/54353373/getting-incorrect-creation-dates-using-aws-s3
+  val clientForCorrectCreatedTime = S3Client
```
akash1810, line 43: Could we combine `client` and `clientForCorrectCreatedTime`? That is, are we able to get all buckets using `Region.US_EAST_1`? A single client would make the code a little simpler.

app/collectors/bucket.scala
```diff
@@ -7,72 +7,96 @@ import conf.AWS
...
+    )
   }
 }
 
 case class Bucket(
   arn: String,
   name: String,
-  region: String,
-  createdTime: Option[Instant]
+  region: Option[String],
```
akash1810, line 98: As you've mentioned, buckets have an almost pseudo region, it's not the most useful information in this regard. Do we know how clients use this property? Or, to ask in a more direct manner, can we afford to drop this property entirely for buckets?

app/collectors/bucket.scala
Nirvikalpa108: Thank you for this great question, Akash. It has made me reflect on the comments I have written and how I can make them clearer. I've made a change to the comment, which states why we cannot use one client. 

The reasoning is that we can only get the correct `createdTime` by using the US-EAST-1 region AND we can only receive data on all the buckets across all the AWS regions, if we use a region other than US-EAST-1 and AWS-GLOBAL. For this reason, I have used two clients.

app/collectors/bucket.scala
Nirvikalpa108: You're right Akash. The region of a bucket has an almost ethereal quality, because it doesn't really matter given that AWS replicate buckets across all available regions. Having said that, I do think it's worth keeping for now, because I'm not sure how other applications are ingesting this field.

akash1810: 👍🏽 worth adding a TODO so we can investigate in later work?

akash1810 (approved):
🚢

app/collectors/bucket.scala
Nirvikalpa108: Good idea!

akash1810 (approved):

app/collectors/bucket.scala
sihil: I think we should retain this. It is shown in the S3 dashboard and is helpful to understand where the canonical location of the data it.

app/collectors/bucket.scala
sihil: Worth saying that this approach is exactly how the AWS S3 web console works. It makes requests to different regions to obtain all of the data and then merges it all together.

app/collectors/bucket.scala
Nirvikalpa108: Thanks for confirming @sihil!👍🏽 Removed TODO in https://github.com/guardian/prism/pull/115

PR #114, opened 2021-01-06
adds RDS Collector

Nirvikalpa108:
## What is the problem?
@paulbrown1982 asked us for information on RDS:

> Hi DevX - Sam our AWS TAM has reminded us that PostGres 9.5 will be going EoL on 16 February. Apparently we have 2 instances running this version in our account. Rather than broadcast I was going to speak to just the respective teams. Does anybody know how I can find out where they are? I can't see rds instances (db.xyz.size) in Prism.

## What is the solution?
Prism should collect data on the Guardian's RDS instances.

## How can we test this?
We tested this on CODE and it's working nicely 🎉
https://prism.code.dev-gutools.co.uk/rds?engine=postgres&engineVersion~=9.5.*


app/jsonimplicits/implicits.scala
```diff
@@ -55,6 +57,7 @@ object model {
     implicit val recurringCharge: Writes[RecurringCharge] = Json.writes[RecurringCharge]
     Json.writes[Reservation]
   }
+  implicit val rdsWriter: Writes[Rds] = Jsonx.formatCaseClass[Rds]
```
sihil, line 60: Might be worth a comment to say why we're not using `Json.writes` here.

app/controllers/Api.scala
```diff
@@ -208,6 +208,13 @@ class Api (cc: ControllerComponents, prismDataStore: Prism, prismConfiguration:
       Api.singleItem(prismDataStore.reservationAgent, arn)
     }
 
+    def rdsList = Action.async { implicit request =>
+      Api.itemList(prismDataStore.rdsAgent, "rds")
```
sihil, line 212: This should be a plural noun for consistency with others. Maybe `rds-instances` or something?

app/jsonimplicits/implicits.scala
Nirvikalpa108, line 60: Scala permits us to have 22 fields in a case class. Our Rds class contained more than that, so we used the `play-json-extensions` library to solve this problem.

conf/routes
```diff
@@ -57,6 +57,9 @@ GET        /data/keys                         controllers.Api.dataKeysList
 GET        /data/lookup/:key                  controllers.Api.dataLookup(key)
 GET        /data/:arn                         controllers.Api.data(arn)
 
+GET        /rds                               controllers.Api.rdsList
+GET        /rds/:arn                          controllers.Api.rds(arn)
```
sihil, line 61: this should match the plural noun used above.

app/collectors/rds.scala
```diff
@@ -0,0 +1,91 @@
...
+    autoMinorVersionUpgrade = db.autoMinorVersionUpgrade,
+    tags = db.tagList.asScala.map(t => t.key -> t.value).toMap,
+  )
+}
+
+case class Rds(
+               arn: String,
+               allocatedStorage: Int,
+               availabilityZone: String,
+               secondaryAvailabilityZone: String,
```
sihil, line 68: This field is currently frequently `null` in CODE so we should make this an `Option[String]`.

sihil (changes requested):

app/collectors/rds.scala
```diff
@@ -0,0 +1,91 @@
...
+    vpcId = db.dbSubnetGroup.vpcId,
+    dbSubnetGroupName = db.dbSubnetGroup.dbSubnetGroupName,
+    vpcSecurityGroupId = db.vpcSecurityGroups.asScala.map(_.vpcSecurityGroupId).toList,
+    storageType = db.storageType,
+    autoMinorVersionUpgrade = db.autoMinorVersionUpgrade,
+    tags = db.tagList.asScala.map(t => t.key -> t.value).toMap,
+  )
+}
+
+case class Rds(
```
Nirvikalpa108, line 64: We decided to omit the following available fields for v1, but we could consider adding these in the future as they may be useful for tooling which connects to DBs.
```
<HostedZoneId>Z2R2ITUGPM61AM</HostedZoneId>
          <Address>mysqldb.########.us-east-1.rds.amazonaws.com</Address>
          <Port>3306</Port>
```
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html

sihil: What is dbInstancePort then? It's always `0`, perhaps we can remove that for now since we don't understand it!

sihil (approved):

Nirvikalpa108:
Looking good on PROD 🎉 https://prism.gutools.co.uk/rds-instances?engine=postgres&engineVersion~=9.5.*

paulbrown1982:
💯  Amazing work!

PR #116, opened 2021-01-07
Make the shelf life for the source agent infinite

sihil:
## What does this change?
The `/sources` endpoint in Prism returns metadata about all of the active collectors. This comes from a combination of configuration read at startup and status data from the individual collectors held in application memory. Unlike a normal collector, it can by definition never be stale. Unfortunately the way that staleness is calculated means that this was stale under some configurations.

This PR modifies the `CrawlRate` object so that the `shelfLife` and `refreshPeriod` fields are `Duration` rather than `FiniteDuration`. This allows them to be set to Duration.Inf and Duration.Undefined in the case of `source` where it cannot become stale and doesn't refresh. 

Note that:
 - If a normal collector is somehow configured with a refresh period that isn't finite we now log a warning and don't schedule collections
 - We also now set the update time of sources to be the boot time of the application.

## How to test
When a configuration that contains low priority regions is deployed, sources no longer becomes stale and AMIgo continues to operate when consuming Prism's sources endpoint.

Low priortity regions can be configured.

### DIFF
--- a/app/agent/collector.scala
+++ b/app/agent/collector.scala
@@ -64,7 +64,7 @@ class CollectorAgent[T<:IndexedItem](val collectorSet: CollectorSet[T], sourceSt
   def init():Unit = {
     log.info(s"Starting agent for collectors: $collectors")
 
-    datumAgents = collectors.map { collector =>
+    datumAgents = collectors.flatMap { collector =>
       val initial = if (lazyStartup) {
         val startupData = Datum.empty[T](collector)
         sourceStatusAgent.update(startupData.label)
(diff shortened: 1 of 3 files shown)


Nirvikalpa108 (approved):
Nice work @sihil and your write-up is very clear too👍🏽

Just to add that this change came about after testing PROD with 6 regions including non high-priority regions and this breaking recipes on Amigo (https://amigo.gutools.co.uk/recipes/amazon-linux-2-arm-docker was returning a 500). 

The stack trace for the error led us to understand that it was prism's /source endpoint that had stale set to true, which was the cause of the problem.

```
Caused by: java.lang.IllegalStateException: AMIgo internal data cache is stale - last update at 2021-01-07T10:04:03.188Z
	at services.PrismAgents$.dataToResult(PrismAgents.scala:27)
	at services.PrismAgents.accounts(PrismAgents.scala:49)
```

Nirvikalpa108:
Merged to PROD with 6 regions in Config and `/sources` is looking good 🎉 The `lastUpdated` time is during the deploy time when config would have been read, which is as expected and the value of stale is false.

```
"sources": [
{
"resource": "sources",
"origin": {
"vendor": "prism",
"accountName": "prism"
},
"status": "success",
"createdAt": "2021-01-07T14:17:38.257Z",
"itemCount": 697,
"age": 527,
"stale": false
}
]
```

PR #121, opened 2021-01-26
Add VPCs to Prism

Nirvikalpa108:
## What does this change?
Here we add a VPC endpoint to Prism.

## Why is this important?

> We have a lot of VPC's at the guardian. These are typically created in two ways:
> - default VPC: comes with the AWS account - generally not recommended!
> - custom VPC: someone has thought about these and had to decide how the subnets in the VPC are split up. Still possibly a suboptimal setup.
> 
> VPCs have a block of IP addresses allocated to them. If set up correctly this will be a block registered with the networking team at the guardian so that one VPC doesn't share IP addresses with another.
> 
> It would be helpful to find out what the current state of play is with all VPCs - which ones are default, and what the IP ranges of each one is. At the same time it makes sense to check the layout of the VPC -e.g. is it the 'preferred' 3 private subnets, 3 public subnets or something different - e.g. dotcom has 3 public 1 private.

https://trello.com/c/ffqKl6oA

Adding VPCs to Prism will give us visibility over the IP ranges of our VPCs and we should be able to correct any that share IP addresses with one another.

## How to test
This is currently on CODE at https://prism.code.dev-gutools.co.uk/vpcs 🎉

### DIFF
--- a/app/collectors/vpc.scala
+++ b/app/collectors/vpc.scala
@@ -46,15 +46,6 @@ object Vpc {
     accountId = vpc.ownerId,
     state = vpc.stateAsString,
     cidrBlock = vpc.cidrBlock,
-    cidrBlockAssociationSet = vpc.cidrBlockAssociationSet.asScala.toList.map{ c =>
-      CidrBlockAssociation(c.cidrBlock, c.associationId, c.cidrBlockState.stateAsString)
-    },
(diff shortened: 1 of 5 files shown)


philmcmahon (approved):
Excellent work 💯

PR #135, opened 2021-02-24
Upgrade to version 0.38.0 of @guardian/cdk

jacobwinch:
## What does this change?

Updates Prism to use `@guardian/cdk` [version 0.38.0](https://github.com/guardian/cdk/releases/tag/v0.38.0). I've commented on the diff below to explain the most significant changes in more detail.

## How to test

I've tested this by deploying the change to `CODE` and confirming that Prism remained accessible throughout the deployment.

* Prism should continue to operate normally during and after the upgrade.

* We're updating security groups and that always feels a bit risky! I think that the testing done mitigates this risk sufficiently though.

### DIFF
--- a/cdk/lib/__snapshots__/prism-access.test.ts.snap
+++ b/cdk/lib/__snapshots__/prism-access.test.ts.snap
@@ -90,7 +90,7 @@ Object {
           },
           Object {
             "Key": "gu:cdk:version",
-            "Value": "0.37.0",
+            "Value": "0.38.0",
           },
           Object {
             "Key": "Stack",
--- a/cdk/lib/prism.ts
+++ b/cdk/lib/prism.ts
@@ -1,5 +1,5 @@
 import { BlockDeviceVolume, EbsDeviceVolumeType, HealthCheck } from "@aws-cdk/aws-autoscaling";
-import { Peer } from "@aws-cdk/aws-ec2";
+import { Peer, Port } from "@aws-cdk/aws-ec2";
 import type { App } from "@aws-cdk/core";
 import { Duration } from "@aws-cdk/core";
 import { Stage } from "@guardian/cdk/lib/constants";
@@ -86,7 +86,7 @@ export class PrismStack extends GuStack {
       ],
     });
 
-    new GuHttpsClassicLoadBalancer(this, "LoadBalancer", {
+    const loadBalancer = new GuHttpsClassicLoadBalancer(this, "LoadBalancer", {
       vpc,
       crossZone: true,
       subnetSelection: { subnets },
@@ -101,5 +101,7 @@ export class PrismStack extends GuStack {
         allowConnectionsFrom: [Peer.ipv4("10.0.0.0/8")],
       },
     });
+
+    appServerSecurityGroup.connections.allowFrom(loadBalancer, Port.tcp(9000), "Port 9000 LB to fleet");
   }
 }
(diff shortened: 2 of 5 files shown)


cdk/lib/__snapshots__/prism-access.test.ts.snap
```diff
@@ -90,7 +90,7 @@ Object {
           },
           Object {
             "Key": "gu:cdk:version",
-            "Value": "0.37.0",
+            "Value": "0.38.0",
```
jacobwinch, line 93: N.B. since this is the only change to this template, I don't intend to update this stack set. Let me know if you think this is necessary.

cdk/lib/prism.ts
```diff
@@ -48,27 +49,33 @@ export class PrismStack extends GuStack {
...
       vpcSubnets: { subnets },
       role: role,
-      userData: userData,
-      minCapacity: 2,
-      maxCapacity: 4,
-      desiredCapacity: 2,
+      userData: userData.userData,
+      stageDependentProps: {
+        [Stage.CODE]: {
+          minimumInstances: 2,
```
jacobwinch, line 69: I wanted to preserve the existing configuration for the upgrade, but perhaps we should run a single instance in `CODE`. I guess it'll save a little bit of money.

cdk/lib/prism.ts
```diff
@@ -94,5 +101,7 @@ export class PrismStack extends GuStack {
         allowConnectionsFrom: [Peer.ipv4("10.0.0.0/8")],
       },
     });
+
+    appServerSecurityGroup.connections.allowFrom(loadBalancer, Port.tcp(9000), "Port 9000 LB to fleet");
```
jacobwinch, line 105: This is necessary to prevent downtime in the period _after_ the CloudFormation update but _before_ the new boxes have been added to the load balancer. For a discussion of similar issues, see https://github.com/guardian/cdk/discussions/212.

We'll be able to remove this in a subsequent PR.

cdk/lib/prism.ts
```diff
@@ -48,27 +49,33 @@ export class PrismStack extends GuStack {
...
+        },
+        [Stage.PROD]: {
+          minimumInstances: 2,
+        },
+      },
       healthCheck: HealthCheck.elb({
         grace: Duration.seconds(500),
       }),
-      securityGroup: appServerSecurityGroup,
+      additionalSecurityGroups: [appServerSecurityGroup],
```
jacobwinch, line 78: We need to keep this for now (we cannot replace a security group in a single CloudFormation update). It's more permissive than the new default security group (see https://github.com/guardian/cdk/pull/268), but I wonder if we can remove it entirely in the future?

cdk/lib/prism.ts
```diff
@@ -48,27 +49,33 @@ export class PrismStack extends GuStack {
...
 
     const distBucket: string = this.getParam(GuDistributionBucketParameter.parameterName).valueAsString;
-    const s3Key = [distBucket, this.stack, this.stage, this.app, `${this.app}.deb`].join("/");
 
-    // TODO move to UserData.forLinux
-    const userData = `#!/bin/bash -ev
-
-      aws --region ${this.region} s3 cp s3://${s3Key} /tmp/
-      dpkg -i /tmp/prism.deb`;
+    const userData = new GuUserData(this, {
```
jacobwinch, line 53: See https://github.com/guardian/cdk/pull/269 for more info on this change.

akash1810 (approved):
Looks good. Keen to discuss security group dancing to see how we can simplify.

cdk/lib/prism.ts
jacobwinch: https://github.com/guardian/prism/pull/137

PR #139, opened 2021-02-26
Migrate to Java 11 runtime

sihil:
## What does this change?
Under ARM64 we are seeing performance issues linked to the Java 8 runtime not having been optimised for the new architecture. This PR migrates prism to Java 11 which does have optimisations for ARM64.

Note that we commonly use a number of command line options that have been removed or deprecated in Java 11.

## How to test
Does prism still work after deploy? Are the GC logs still there?

Might be worth looking at average CPU usage before and after deployment.

 * This might be unreliable as we've not really used Java 11 anywhere. 
 * We've translated JVM parameters from 8 to 11, the translation might be wrong.
 * There is a new AMIgo role which might have errors.

That said Java 11 is widely used and we are still compiling under JDK8 so this should be low risk.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,11 +78,9 @@ lazy val root = (project in file("."))
 
 javaOptions in Universal ++= Seq(
   "-Dpidfile.path=/dev/null",
-  "-J-XX:MaxRAMFraction=2",
-  "-J-XX:InitialRAMFraction=2",
+  "-J-XX:MaxRAMPercentage=60",
+  "-J-XX:InitialRAMPercentage=60",
   "-J-XX:MaxMetaspaceSize=300m",
-  "-J-XX:+PrintGCDetails",
-  "-J-XX:+PrintGCDateStamps",
-  s"-J-Xloggc:/var/log/${packageName.value}/gc.log"
+  s"-J-Xlog:gc*:file=/var/log/${packageName.value}/gc.log:time:filecount=5,filesize=1024K"
 )
 
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -9,7 +9,7 @@ deployments:
       amiParameter: AMI
       amiEncrypted: true
       amiTags:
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -78,11 +78,9 @@ lazy val root = (project in file("."))
 
 javaOptions in Universal ++= Seq(
   "-Dpidfile.path=/dev/null",
-  "-J-XX:MaxRAMFraction=2",
-  "-J-XX:InitialRAMFraction=2",
+  "-J-XX:MaxRAMPercentage=60",
+  "-J-XX:InitialRAMPercentage=60",
```
sihil, line 82: Rather than 50%, I've moved to 60% - this was desired but never possible under the previous setup. We should consider making this even higher now that we can!

build.sbt
```diff
@@ -78,11 +78,9 @@ lazy val root = (project in file("."))
...
   "-J-XX:MaxMetaspaceSize=300m",
-  "-J-XX:+PrintGCDetails",
-  "-J-XX:+PrintGCDateStamps",
-  s"-J-Xloggc:/var/log/${packageName.value}/gc.log"
+  s"-J-Xlog:gc*:file=/var/log/${packageName.value}/gc.log:time:filecount=5,filesize=1024K"
```
sihil, line 84: This line has various components that replaces that of the previous three lines.

akash1810 (approved):

build.sbt
philmcmahon: Do you mean now that we can because of Java 11 or because we have more RAM on our machines?

johnduffell: that's getting risky, aas the 60% is only the heap, and there is something else that it also limits to exactly the same amount, meaning that the JVM can use all the system ram plus more.
I found that information when investigating a system out of memory on support-frontend after it had been running a while.

johnduffell: see https://github.com/guardian/support-frontend/pull/2907

PR #155, opened 2021-03-17
chore(deps): upgrade @guardian/cdk to latest (8.2.0)

akash1810:
## What does this change?

Updates to [@guardian/cdk v8.2.0](https://github.com/guardian/cdk/releases/tag/v8.2.0) 🎉 . This has got a number of breaking changes (full diff [here](https://github.com/guardian/cdk/compare/v0.38.0...v8.2.0)).

We'd need to manually apply the CFN template as the following parameters have been renamed (hopefully for the last time!):
- AMI -> AMIPrism
- InstanceType -> InstanceTypePrism

## How to test

n/a

Keeping up to date.

n/a

## Images

n/a

### DIFF
--- a/cdk/lib/__snapshots__/prism.test.ts.snap
+++ b/cdk/lib/__snapshots__/prism.test.ts.snap
@@ -127,7 +127,7 @@ Object {
         },
         "LoadBalancerNames": Array [
           Object {
-            "Ref": "LoadBalancerBE9EEC3A",
+            "Ref": "LoadBalancer",
           },
         ],
         "MaxSize": Object {
@@ -573,7 +573,7 @@ dpkg -i /prism/prism.deb",
       },
       "Type": "AWS::IAM::Role",
     },
-    "LoadBalancerBE9EEC3A": Object {
+    "LoadBalancer": Object {
       "Properties": Object {
         "CrossZone": true,
         "HealthCheck": Object {
--- a/cdk/lib/__snapshots__/prism.test.ts.snap
+++ b/cdk/lib/__snapshots__/prism.test.ts.snap
@@ -15,8 +15,8 @@ Object {
     },
   },
   "Parameters": Object {
-    "AMI": Object {
-      "Description": "AMI ID",
+    "AMIPrism": Object {
+      "Description": "Amazon Machine Image ID for the app prism. Use this in conjunction with AMIgo to keep AMIs up to date.",
       "Type": "AWS::EC2::Image::Id",
     },
     "CertificateARN": Object {
@@ -30,9 +30,9 @@ Object {
       "Description": "SSM parameter containing the S3 bucket name holding distribution artifacts",
       "Type": "AWS::SSM::Parameter::Value<String>",
     },
-    "InstanceType": Object {
+    "InstanceTypePrism": Object {
       "Default": "t3.small",
-      "Description": "EC2 Instance Type",
+      "Description": "EC2 Instance Type for the app prism",
       "Type": "String",
     },
     "LoggingStreamName": Object {
@@ -78,7 +78,7 @@ Object {
           },
           Object {
             "Key": "gu:cdk:version",
-            "Value": "0.38.0",
+            "Value": "8.2.0",
           },
           Object {
             "Key": "Stack",
@@ -97,7 +97,7 @@ Object {
       },
       "Type": "AWS::EC2::SecurityGroup",
     },
-    "AppServerSecurityGroupfromprismLoadBalancerSecurityGroupA5954B5A90003B574E9E": Object {
+    "AppServerSecurityGroupPrismfromprismLoadBalancerSecurityGroupA5954B5A90002D5FC675": Object {
       "Properties": Object {
         "Description": "Port 9000 LB to fleet",
         "FromPort": 9000,
@@ -123,11 +123,11 @@ Object {
         "HealthCheckGracePeriod": 500,
         "HealthCheckType": "ELB",
         "LaunchConfigurationName": Object {
-          "Ref": "AutoscalingGroupLaunchConfigEBD75746",
+          "Ref": "AutoscalingGroupPrismLaunchConfig69D33207",
         },
         "LoadBalancerNames": Array [
           Object {
-            "Ref": "LoadBalancer",
+            "Ref": "LoadBalancerBE9EEC3A",
           },
         ],
         "MaxSize": Object {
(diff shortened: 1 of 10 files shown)


cdk/lib/__snapshots__/prism.test.ts.snap
```diff
@@ -642,13 +625,9 @@ dpkg -i /prism/prism.deb",
           },
         ],
         "Tags": Array [
-          Object {
```
jacobwinch, line 645: A few resources are losing their `App` tag in this PR - should we be concerned about that?

jacobwinch (dismissed):

cdk/lib/__snapshots__/prism.test.ts.snap
akash1810: Yes we should be, thanks for flagging this! It looks like more constructs need to take in the `app` prop than were updated in https://github.com/guardian/cdk/pull/326. We'll need to fix this before merging this in.

akash1810: Making this PR draft until @guardian/cdk has been updated to fix this.

akash1810: https://github.com/guardian/cdk/pull/378 should solve this.

cdk/lib/prism-access.ts
```diff
@@ -3,20 +3,32 @@ import type { App } from "@aws-cdk/core";
...
   constructor(scope: App, id: string, props: GuStackProps) {
     super(scope, id, props);
 
+    /*
+    Looks like some @guardian/cdk constructs are not applying the App tag.
+    I suspect since https://github.com/guardian/cdk/pull/326.
+    Until that is fixed, we can safely, manually apply it to all constructs in tree from `this` as it's a single app stack.
+    TODO: remove this once @guardian/cdk has been fixed.
+     */
+    AppIdentity.taggedConstruct(PrismAccess.app, this);
```
akash1810, line 23: This can be seen in the snapshot diff on the [first commit](https://github.com/guardian/prism/pull/155/commits/dc69fbee724965a49d9db54ef356fb18867e368c).

cdk/lib/__snapshots__/prism.test.ts.snap
```diff
@@ -573,7 +573,7 @@ dpkg -i /prism/prism.deb",
       },
       "Type": "AWS::IAM::Role",
     },
-    "LoadBalancer": Object {
+    "LoadBalancerBE9EEC3A": Object {
```
jacobwinch, line 576: @akash1810 - I think we'll need to set `existingLogicalId` on the load balancer. I'm a bit confused as to how this id has remained static so far, as `overrideId` doesn't currently appear to be set 😕

cdk/lib/__snapshots__/prism.test.ts.snap
```diff
@@ -415,22 +415,22 @@ dpkg -i /prism/prism.deb",
...
           },
         ],
       },
       "Type": "AWS::IAM::Policy",
     },
-    "GuHttpsEgressSecurityGroupF63CDA96": Object {
+    "GuHttpsEgressSecurityGroupPrism4A68FA56": Object {
       "Properties": Object {
-        "GroupDescription": "Allow all outbound traffic on port 443",
+        "GroupDescription": "Allow all outbound HTTPS traffic",
```
akash1810, line 429: This will result in a replacement 😢 . I wonder if guardian/cdk could send a warning on this event? Though maybe the major version number change is enough of a signal?

cdk/lib/__snapshots__/prism.test.ts.snap
akash1810: Argh! Good spot! Will add `existingLogicalId`.

I think this is the result of https://github.com/guardian/cdk/pull/440. Looking at the [removed tests](https://github.com/guardian/cdk/commit/0d0eba739e97c18baa411f3f4eb58b6fa26727bb#diff-ddcd012c9ad844faf4c3b865a658f5e3e600d62ad88228ca022798def7c56c9dL33), it looks like if `migratedFromCloudFormation` was true on the `GuStack`, the construct's logicalId was not auto-generated; the value passed into the constructor was used.

I think this is another +1 for https://github.com/guardian/cdk/pull/364 and encapsulate this logic in one place.

I'd also have hoped to see [the message](https://github.com/guardian/cdk/blob/3eefaaab116b101498151ab8b9e94dc996d9eb21/src/constructs/core/migrating.ts#L70) in the console, but then I realised I've been using Jest locally which sets NODE_ENV to test - oops https://github.com/guardian/cdk/pull/397!

cdk/lib/__snapshots__/prism.test.ts.snap
jacobwinch: Hmm, good point. We have (at least) two different types of breaking change:

1. We've changed the API and you'll need to update some code, but it shouldn't cause any breaking changes to the underlying infrastructure 
2. We've changed something which will have a significant impact on the underlying infrastructure (e.g. replacing a security group, updating load balancer version)

I think everyone will be familiar with type 1, but I wonder if a major version bump will be enough to prepare them for changes of type 2... hopefully these will be rare events once we have our production-ready version though!

jacobwinch (approved):
Great work 💯

akash1810:
This change updates security groups. This is normally tricky when they are in-use by instances as the instances need to be cycled out of service.

This manifests itself like this in the CFN event log:

> resource sg-0af9f2fea82e07d81 has a dependent object (Service: AmazonEC2; Status Code: 400; Error Code: DependencyViolation; Request ID: af537518-65ca-454c-b652-e2b1d11ca2c9; Proxy: null)

When this happens, I usually perform an app only deploy to cycle the current instances out and launch new ones.

I think I did this too early on CODE, which lead to a few issues. That is, I didn't wait for the `DependencyViolation` message to  perform an app only deploy, instead I deployed earlier. This resulted in new instances failing to launch as they couldn't get the deb file from S3. It is most likely that the IAM policy changes hadn't finished.

I've deployed the CFN change manually to PROD, waiting for the `DependencyViolation` event to cycle the instances out. This went without issue.

jacobwinch:
> I think I did this too early on CODE, which lead to a few issues. That is, I didn't wait for the `DependencyViolation` message to perform an app only deploy, instead I deployed earlier at a point when it looks like the IAM policy changes hadn't finished.

Ah this makes sense now! Thanks for explaining!

PR #168, opened 2021-03-31
Add new route for monitoring cdk usage

jacobwinch:
## What does this change?

1. Adds a new `/apps-with-cdk-version` route, so that we can track adoption/upgrade status of https://github.com/guardian/cdk across our estate
2. Updates the `/apps` response so that lambda-based apps are also included (this was previously reporting EC2-based apps only, which I think is a bug/oversight)
3. Adds `app` information to `/lambdas` response (where relevant)
4. Adds `guCdkVersion` information to `/instances` and `/lambdas` responses (where relevant)

In order to support 1, I considered an alternative implementation based on querying all stacks via the CloudFormation API. I decided not to pursue this as it would have required new IAM permissions and scraping of additional AWS data. It would also likely have been harder/slower to implement than the implementation proposed in this PR!

## How to test

This change is on `CODE` at the time of writing and seems to be working as expected.

It's possible to find out which projects have started using `guardian/cdk`. In the future we'll also be able to use this to identify projects running with older versions of the library.

I think this should be safe as we are only adding new responses / adding more fields to existing responses. We are not fetching any new data from AWS to support this functionality so it should not significantly affect performance. 

There is a small risk that a client is relying on the EC2-only nature of the current `/apps` response, but I think this is unlikely.

### DIFF
--- a/app/collectors/lambda.scala
+++ b/app/collectors/lambda.scala
@@ -63,6 +63,8 @@ object Lambda extends Logging{
     region,
     runtime = getRuntime(lambda),
     tags,
+    app = tags.get("App").map(_.split(",").toList).getOrElse(Nil),
+    guCdkVersion = tags.get("gu:cdk:version"),
     stage = tags.get("Stage"),
     stack = tags.get("Stack")
   )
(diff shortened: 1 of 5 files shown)


app/collectors/lambda.scala
```diff
@@ -74,8 +76,10 @@ case class Lambda(
   region: String,
   runtime: String,
   tags: Map[String, String],
+  override val app: List[String],
```
jacobwinch, line 79: I have used `List[String]` (rather than `Option[String]`) for consistency with the `instance` model. 

I suspect that this choice of type may be a legacy thing, so I briefly attempted to refactor and use `Option[String]` for both. However, we are already exposing this field to clients as an array (e.g. `"app":["mobile-sport"]`), so I don't think we can easily update this type.

philmcmahon (approved):
LGTM nice work

akash1810 (commented):
Awesome work!

PR #197, opened 2021-05-07
Crawl cloudformation stacks and resources

sihil:
## What does this change?
There is a rich seam of information in the cloudformation stacks. This extracts it every few minutes and makes it available to prism clients. With this we can more easily ask questions about resources that are or are not looked after by cloudformation and identify the stack that a resource belonds to more quickly.

## How to test
This should be tested in CODE to check that is works as expected and doesn't reveal private data.

 - Being able to identify all S3 buckets not created in Cloudformation.
 - Being able to identify any CFN stacks that have stack policies

There are a few risks worth a little more thinking about:
 - will the extra data make prism too slow again?
 - is any of the data in outputs and parameters for stacks be private?

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -223,6 +223,14 @@ class Api (cc: ControllerComponents, prismDataStore: Prism, prismConfiguration: 
       Api.singleItem(prismDataStore.vpcAgent, arn)
     }
 
+    def cloudformationStackList = Action.async { implicit request =>
+      Api.itemList(prismDataStore.cloudformationStackAgent, "cloudformation-stacks")
+    }
+
+    def cloudformationStack(arn: String) = Action.async { implicit request =>
+      Api.singleItem(prismDataStore.cloudformationStackAgent, arn)
+    }
+
     private def stackExtractor(i: IndexedItemWithStack) = i.stack.map(Json.toJson(_))
     private def stageExtractor(i: IndexedItemWithStage) = i.stage.map(Json.toJson(_))
     def roleList = summary[Instance](prismDataStore.instanceAgent, i => i.role.map(Json.toJson(_)), "roles")
--- a/app/collectors/cloudformationStacks.scala
+++ b/app/collectors/cloudformationStacks.scala
@@ -69,8 +69,11 @@ object CloudformationStack {
         CloudformationStackOutput(
           description = Option(output.description),
           exportName = Option(output.exportName),
-          key = Option(output.outputKey),
-          value = Option(output.outputValue)
+          key = None,
+          value = None,
+          // commented out whilst we figure out whether it's safe to display all values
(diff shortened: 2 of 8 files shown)


app/utils/filter.scala
```diff
@@ -71,5 +77,8 @@ object ResourceFilter {
     ResourceFilter((defaultKeys ++ filterKeys).toMap)
   }
 
-  lazy val all: Matchable[JsValue] = (_: JsValue) => true
+  lazy val all: Matchable[JsValue] = new Matchable[JsValue] {
+    override def isMatch(value: JsValue): Boolean = true
+    override def inverse: Boolean = false
+  }
```
akash1810, line 83: What are these changes for?

akash1810 (commented):
The change set is a lot smaller than I imagined - its way too simple to crawl a new AWS resource!

I don't fully understand the changes to `filter.scala`

app/utils/filter.scala
sihil: ah - I knew there was something I still needed to test. I'll document it and ping you.

sihil: I've now extracted the filtering changes into their own PR #214 for clarity and ability to independently roll either back.

akash1810 (approved):
LGTM!

app/controllers/Api.scala
```diff
@@ -223,6 +223,14 @@ class Api (cc: ControllerComponents, prismDataStore: Prism, prismConfiguration:
       Api.singleItem(prismDataStore.vpcAgent, arn)
     }
 
+    def cloudformationStackList = Action.async { implicit request =>
+      Api.itemList(prismDataStore.cloudformationStackAgent, "cloudformation-stacks")
```
akash1810, line 227: Spotting whilst testing on CODE - kebab casing here makes the `jq` query a bit annoying as you have to [wrap the field in quotes](https://stackoverflow.com/a/37344498/3868241):

`curl https://prism.../cloudformationStacks | jq '.data."cloudformation-stacks"[].outputs'`

Could this be `cloudformationStacks` instead?

`curl https://prism.../cloudformationStacks | jq '.data.cloudformationStacks[].outputs'`

Not a blocker.

akash1810:
Shipping this as @sihil is on paternity leave.

- [ ] Update StackSet to add the additional permissions
- [ ] Deploy to CODE
- [ ] Check that no stack outputs secrets
- [ ] Deploy to PROD

akash1810 (approved):

akash1810:
> is any of the data in outputs and parameters for stacks be private?

It's worth noting this PR explicitly sets these values to `None` in the response so we can ship this now whilst we get teams to update their stacks.

akash1810:
I've deployed this branch to PROD w/out issues. Merging.

PR #212, opened 2021-05-25
Ensure that logical ID of prism role doesn't change

sihil:
## What does this change?
During an update the logical ID of the prism access role was modified. This overrides it to ensure that the roles are not replaced, which would be hours of bad.

## How to test
Trial this on a single account to ensure it doesn't replace the role.


cdk/lib/prism-access.ts
```diff
@@ -29,6 +29,7 @@ export class PrismAccess extends GuStack {
     };
 
     const prismRole = new GuRole(this, "PrismRole", {
+      existingLogicalId: "PrismRole",
```
akash1810, line 32: nitpick: worth adding a comment here to explain why the default id is undesired?

This prop should really be changed to enforce this:

```typescript
interface ExistingLogical {
  value: string;
  description: string;
}
```

akash1810 (approved):

akash1810 (approved):
👍🏽 

Hopefully the additional comments will prevent any future disasters!

PR #214, opened 2021-05-28
Add ability to filter by existence of fields

sihil:
## What does this change?
One of the queries that was impossible to do without leaning on `jq` was to find records where a field was not defined. This was mainly down to weird logic where the inversion was done after the type of value was discovered and any time the valuewas null we would never match the object regardless of any filters. This improves that situation by rearranging the structure of the logic so inversion is done before the type of value is assessed.

This also fixes a longstanding issue with not being able to search on boolean fields for the same reason as above.

## How to test
There are new tests added but you can test with any boolean fields or by doing existance queries. For example: `https://prism.gutools.co.uk/lambdas?runtime!=` now works as you might expect.

Easier to identify resources by wider criteria.

It's possible that existing corner case behaviour has changed a little which might break assumptions used elsewhere in consumers of Prism. This is worth watching for and rolling back if need be. Most likely riff-raff or amiable.

### DIFF
--- a/test/ApiSpec.scala
+++ b/test/ApiSpec.scala
@@ -67,7 +67,7 @@ object ApiSpec extends PlaySpecification with Results {
     }
   }
 
-  case class TestItem(arn: String, name: String, region: String, tags: Map[String, String] = Map.empty) extends IndexedItem {
+  case class TestItem(arn: String, name: String, region: String, tags: Map[String, String] = Map.empty, flag: Boolean = false, optionalMap: Option[Map[String, String]] = None) extends IndexedItem {
     def callFromArn: String => Call = _ => Call("GET", "localhost")
   }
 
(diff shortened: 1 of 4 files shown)


test/ApiSpec.scala
```diff
@@ -87,22 +87,27 @@ object ApiSpec extends PlaySpecification with Results {
...
+    TestItem("arn", "name", "eu-west-2", Map("stage" -> "CODE"), flag=true)
+  )
 
-  class TestCollectorAgent extends CollectorAgentTrait[TestItem] {
+  class TestCollectorAgent(items: Seq[TestItem] = defaultItems) extends CollectorAgentTrait[TestItem] {
     private val resourceType = ResourceType("test")
     private val label = Label(resourceType, new TestOrigin, 1)
 
-    def get(collector: Collector[TestItem]): Datum[TestItem] = Datum(label, Seq(TestItem("arn", "name", "region")))
+    def get(collector: Collector[TestItem]): Datum[TestItem] = ???
```
akash1810, line 100: Not sure if these need an implementation as we're in tests here... If we do not need an implementation, can we add a comment to explain why?

sihil:
@akash1810 Not sure if you want to merge this or close. I'm tidying up all my old PRs so my GH wide PR dashboard is useful so I'll close this and other similar ones in the coming days unless flagged otherwise :)

kenoir:
> @akash1810 Not sure if you want to merge this or close. I'm tidying up all my old PRs so my GH wide PR dashboard is useful so I'll close this and other similar ones in the coming days unless flagged otherwise :)

Closing as out of date.

PR #223, opened 2021-06-07
feat(cdk): bump @guardian/cdk to v19

stephengeller:
## What does this change?

This PR bumps the `@guardian/cdk` to [v19.0.0](https://github.com/guardian/cdk/releases/tag/v19.0.0), allowing us to eventually make use of the [v18.1.0's new pattern tag](https://github.com/guardian/cdk/pull/603).

Once we migrate to the GuEc2App pattern, this will allow us to get the first piece of data of pattern usage into AWS, allowing us to build a lambda to query this data and offload it to CloudWatch and later Grafana.

## How to test

I've manually applied this CFN update in `CODE` and confirmed that it doesn't cause any problems.

Prism's ASG is tagged with what pattern uses it.

### DIFF
--- a/cdk/lib/__snapshots__/prism.test.ts.snap
+++ b/cdk/lib/__snapshots__/prism.test.ts.snap
@@ -81,6 +81,10 @@ Object {
             "Value": "TEST",
           },
           Object {
+            "Key": "gu:repo",
+            "Value": "guardian/prism",
+          },
+          Object {
             "Key": "Stack",
             "Value": "deploy",
           },
@@ -97,7 +101,7 @@ Object {
       },
       "Type": "AWS::EC2::SecurityGroup",
     },
-    "AppServerSecurityGroupPrismfromprismLoadBalancerSecurityGroupA5954B5A90002D5FC675": Object {
+    "AppServerSecurityGroupPrismfromprismLoadBalancerPrismSecurityGroupE6A4FDFF900028CF9C72": Object {
       "Properties": Object {
         "Description": "Port 9000 LB to fleet",
         "FromPort": 9000,
@@ -110,7 +114,7 @@ Object {
         "IpProtocol": "tcp",
         "SourceSecurityGroupId": Object {
           "Fn::GetAtt": Array [
-            "LoadBalancerSecurityGroupA28D6DD7",
+            "LoadBalancerPrismSecurityGroupB966F197",
             "GroupId",
           ],
         },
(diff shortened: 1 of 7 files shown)


cdk/lib/__snapshots__/prism.test.ts.snap
```diff
@@ -785,6 +816,55 @@ dpkg -i /prism/prism.deb",
...
+        ],
+        "VpcId": Object {
+          "Ref": "VpcId",
+        },
+      },
+      "Type": "AWS::EC2::SecurityGroup",
+    },
   },
 }
 `;
```
jacobwinch, line 870: It looks like there are quite a few changes in this file which will cause security groups to be replaced (updating logical ids and descriptions). We'll need to figure out how to bring new instances up at the right time to avoid the CFN update hanging or downtime (@akash1810 has a neat trick for this, but I'm struggling to remember it!). If you haven't already, you might want to rehearse this process in `CODE` to ensure that it goes smoothly before merging this in.

akash1810: The dance I do is:
1. Apply the CFN manually (either in the web console or with `cdk deploy`)
2. Wait until the CFN update deploy "fails" with a failure to delete an in-use security group
3. Perform an autoscaling deploy using RiffRaff. This will cycle the instances using the above SG out of service allowing the SG to be deleted
4. Wait for the CFN update to finish. You should now be in a fully working state w/no downtime.
5. Perform a full deploy with RIffRaff to ensure CD would be ok

Depending on how it's timed, you sometimes have to do 2 and 3 a couple of times.

cdk/lib/__snapshots__/prism.test.ts.snap
```diff
@@ -785,6 +816,55 @@ dpkg -i /prism/prism.deb",
       },
       "Type": "AWS::IAM::Policy",
     },
+    "WazuhSecurityGroup": Object {
```
akash1810, line 819: I'm a little surprised we didn't have Wazuh already. IIUC the agent is installed on the AMI by default in AMIgo now.

cc @jorgeazevedo @philmcmahon @Nirvikalpa108

jacobwinch (approved):

cdk/lib/__snapshots__/prism.test.ts.snap
jacobwinch: In this case the update can actually be applied cleanly without this dance. Although it looks like we are adding/deleting security groups, upon closer inspection I can see that we are predominantly adding/deleting resources of type `AWS::EC2::SecurityGroupIngress`/`AWS::EC2::SecurityGroupEgress`. The only `AWS::EC2::SecurityGroup` which gets replaced is attached to the LB (not the instances)/ AWS seems to be able to apply all of these changes cleanly in one update without introducing downtime.

PR #227, opened 2021-06-15
Increase ASG healthcheck grace period (again)

jacobwinch:
This is a more generous version of https://github.com/guardian/prism/pull/226, as 600 seconds still doesn't seem to be sufficient! This [dashboard](https://logs.gutools.co.uk/s/devx/goto/62844fe86e35160bcb41ebcfa1ebd44a) suggests that 800 seconds should be OK.

### DIFF
--- a/cdk/lib/prism.ts
+++ b/cdk/lib/prism.ts
@@ -87,7 +87,7 @@ export class PrismStack extends GuStack {
         },
       },
       healthCheck: HealthCheck.elb({
-        grace: Duration.seconds(600),
+        grace: Duration.seconds(800),
       }),
       additionalSecurityGroups: [appServerSecurityGroup],
       blockDevices: [
--- a/cdk/lib/__snapshots__/prism.test.ts.snap
+++ b/cdk/lib/__snapshots__/prism.test.ts.snap
@@ -120,7 +120,7 @@ Object {
     },
     "AutoscalingGroup": Object {
       "Properties": Object {
-        "HealthCheckGracePeriod": 600,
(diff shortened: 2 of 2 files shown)


cdk/lib/prism.ts
```diff
@@ -87,7 +87,7 @@ export class PrismStack extends GuStack {
         },
       },
       healthCheck: HealthCheck.elb({
-        grace: Duration.seconds(600),
+        grace: Duration.seconds(800),
```
akash1810, line 90: Could we use [`Duration.minutes`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Duration.html#static-minutesamount) here instead?

jacobwinch: The docs, CFN, console all use seconds so I'm not sure it helps us that much to use a different unit here, especially with 800 which doesn't convert into a whole number of minutes (although admittedly 800 was a slightly arbitrary choice!)

akash1810: Ah! `Duration.minutes(8).toSeconds()` would work? Not massively more readable though esp if the duration isn't divisible by 60.

akash1810 (approved):

cdk/lib/prism.ts
jacobwinch: You can do `Duration.minutes(13.3)` and `cdk` will convert it to 798 seconds, but I don't think this really makes things better as the console will still show 798 (and I'll still end up having to Google the conversion 😁)!

PR #230, opened 2021-06-24
Upgrade @guardian/cdk to v19.2.1

jacobwinch:
## What does this change?

Upgrades to use @guardian/cdk to v19.2.1. We are already on v19.0.0, so the changes here are minimal. The primary motivation for this is to pick up https://github.com/guardian/cdk/pull/609, which is required for the migration to the EC2 App Pattern.

## How to test

The snapshot tests show that the generated template is unchanged. This should be sufficient.

We will need to make fewer changes in the EC2 App Pattern migration PR(s).

This seems very low risk to me.

### DIFF
--- a/cdk/lib/prism.ts
+++ b/cdk/lib/prism.ts
@@ -77,7 +77,8 @@ export class PrismStack extends GuStack {
       ...PrismStack.app,
       existingLogicalId: {
         logicalId: "AutoscalingGroup",
-        reason: "We override this to ensure that we do not replace the existing resource",
+        reason:
+          "We override this to ensure that we do not replace the existing resource (as this may result in downtime)",
       },
       vpc,
       vpcSubnets: { subnets },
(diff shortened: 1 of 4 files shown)


cdk/lib/prism.ts
```diff
@@ -72,7 +75,10 @@ export class PrismStack extends GuStack {
 
     const asg = new GuAutoScalingGroup(this, "AutoscalingGroup", {
       ...PrismStack.app,
-      existingLogicalId: "AutoscalingGroup",
+      existingLogicalId: {
+        logicalId: "AutoscalingGroup",
+        reason: "We override this to ensure that we do not replace the existing resource",
```
akash1810, line 80: Worth being a little more descriptive? e.g. is this stateful and replacing it will result in downtime?

jacobwinch: I had to pause whilst copy & pasting this one... I'm not actually completely sure if replacing an ASG is a problem? I guess it depends whether AWS waits for the new ASG's boxes to come into service before deleting the old one...

stephengeller (approved):

PR #232, opened 2021-06-29
Add VPCS to readme entities list. Linkify urls

philmcmahon:
## What does this change?
Prism scans VPCs too! Also, linkify all the entities so people can jump straight in


jorgeazevedo (approved):
ah, nice! very convenient

jonathonherbert (approved):

PR #245, opened 2021-07-07
Generate a separate stack for Prism (using cdk pattern)

jacobwinch:
## What does this change?

This PR creates a new CloudFormation stack*, which defines Prism's infrastructure using our `cdk` pattern. This PR does not alter (or tear down) the current stack, as this must happen via a multi-stage process. At the end of the process, the most significant change to the underlying infrastructure will be our usage of an ALB instead of an ELB.

In this PR we are creating a new load balancer that will run in parallel to the current one. This will allow us to perform a zero-downtime migration via a three step dance:

1. Run two load balancers
1. Update the CNAME to point to the new load balancer
1. Remove the old load balancer

Riff-Raff will not be capable of deploying the stack created via this PR (it must be CloudFormed manually). A subsequent PR (https://github.com/guardian/prism/pull/244), which must be merged **after** the aforementioned DNS change, will remove the old version of the infrastructure and re-enable CFN deployments via Riff-Raff (more details on the migration strategy can be found below).

## How to test

I've deployed the generated template in `CODE` and confirmed that:

- Data is still being returned as expected for common URLs
- The new load balancer is reachable on VPN / not reachable when disconnected from VPN
- Instances in the new stack are still capable of logging to Central ELK

I didn't want to remove the current `CODE` stack until the suggested process has been peer-reviewed, so I simulated its removal by hacking the tags on original stack and ASG, which is sufficient to make Riff-Raff ignore them! This allowed me to confirm that the changes in https://github.com/guardian/prism/pull/244 are sufficient to allow:

- Riff-Raff to make CFN changes to the new stack 
- Riff-Raff to update the new ASG

At the end of the migration process described above, Prism will be using our pattern (which is the recommended approach to using `@guardian/cdk`).

## Migration strategy for PROD

This migration path is not ideal**; deployments will be blocked during the migration. However, as Prism is rarely worked on outside of Developer Experience I think this is tolerable as long as we complete the process relatively quickly.

Here are the suggested steps:

1. Reduce TTL for DNS
    _* This allows us to roll back faster if something goes wrong_
1. Impose deployment restriction in Riff-Raff 
    _* This prevents accidental changes being made during the migration_
1. Merge this PR
1. Run `./scripts/ci` to generate the CloudFormation file
1. Manually create new CloudFormation stack with appropriate tags
1. Simulate a DNS change (by editing `/etc/hosts`) and confirm that the new version of the `PROD` infrastructure is functional
1. Update DNS and allow TTL to expire
1. Manually scale down old ASG and confirm that everything is still working as expected
    _* This allows us to recover more quickly than if we had deleted the whole stack_
1. ~~Remove Deletion Protection from the old load balancer~~ Edit: this is only required for ALBs and we're removing an ELB.
1. Manually remove the old CloudFormation stack via the AWS console
1. Reinstate original TTL
1. Update https://github.com/guardian/prism/pull/244 to point at the correct branch and merge it in
1. Remove deployment restriction in Riff-Raff
1. Manually deploy latest version of Prism, from `main` 
1. Ensure Riff-Raff deployment works correctly
1. Deploy a small infrastructure change and ensure that it works correctly

Yes, the migration process proposes is complex and involves many manual steps. I've tried to incorporate some safety measures to reduce risk (see _explanations in italics_ above).

*conveniently the current stacks are called `Prism-<STAGE>` and the more common convention in the account seems to be `prism-<STAGE>`, so I can fix that too!
**I spent a long time experimenting with different migration strategies, but was not able to get them working with Prism

### DIFF
--- a/cdk/lib/prism-ec2-app.ts
+++ b/cdk/lib/prism-ec2-app.ts
@@ -1,0 +1,80 @@
+import type { CfnAutoScalingGroup } from "@aws-cdk/aws-autoscaling";
+import { BlockDeviceVolume, EbsDeviceVolumeType } from "@aws-cdk/aws-autoscaling";
+import { Peer } from "@aws-cdk/aws-ec2";
+import type { App } from "@aws-cdk/core";
+import { Duration } from "@aws-cdk/core";
+import { AccessScope, GuPlayApp } from "@guardian/cdk";
+import type { AppIdentity } from "@guardian/cdk/lib/constructs/core/identity";
+import type { GuStackProps } from "@guardian/cdk/lib/constructs/core/stack";
+import { GuStack } from "@guardian/cdk/lib/constructs/core/stack";
+import {
+  GuAllowPolicy,
+  GuAssumeRolePolicy,
+  GuDynamoDBReadPolicy,
+  GuGetS3ObjectsPolicy,
+} from "@guardian/cdk/lib/constructs/iam";
+
+export class PrismEc2App extends GuStack {
+  private static app: AppIdentity = {
+    app: "prism",
+  };
+
+  constructor(scope: App, id: string, props: GuStackProps) {
+    super(scope, id, props);
+
+    const pattern = new GuPlayApp(this, {
+      ...PrismEc2App.app,
+      userData: {
+        distributable: {
+          fileName: "prism.deb",
+          executionStatement: `dpkg -i /${PrismEc2App.app.app}/prism.deb`,
+        },
+      },
+      certificateProps: {
+        CODE: { domainName: "prism.code.dev-gutools.co.uk" },
+        PROD: { domainName: "prism.gutools.co.uk" },
+      },
+      monitoringConfiguration: { noMonitoring: true },
+      access: { scope: AccessScope.INTERNAL, cidrRanges: [Peer.ipv4("10.0.0.0/8")] },
+      roleConfiguration: {
+        additionalPolicies: [
+          new GuAllowPolicy(this, "DescribeEC2BonusPolicy", {
+            resources: ["*"],
+            actions: ["EC2:Describe*"],
+          }),
+          new GuDynamoDBReadPolicy(this, "ConfigPolicy", { tableName: "config-deploy" }),
+          new GuGetS3ObjectsPolicy(this, "DataPolicy", {
+            bucketName: "prism-data",
+          }),
+          new GuAssumeRolePolicy(this, "CrawlerPolicy", {
+            resources: ["arn:aws:iam::*:role/*Prism*", "arn:aws:iam::*:role/*prism*"],
+          }),
+        ],
+      },
+      scaling: {
+        CODE: { minimumInstances: 1 },
+        PROD: { minimumInstances: 2 },
+      },
+      blockDevices: [
+        {
+          deviceName: "/dev/sda1",
+          volume: BlockDeviceVolume.ebs(8, {
+            volumeType: EbsDeviceVolumeType.GP2,
+          }),
+        },
+      ],
+    });
+
+    // The pattern does not currently offer support for customising healthchecks via props
+    pattern.targetGroup.configureHealthCheck({
+      path: "/management/healthcheck",
+      unhealthyThresholdCount: 10,
+      interval: Duration.seconds(5),
+      timeout: Duration.seconds(3),
+    });
+
+    // Similarly the pattern does not offer support for extending the default ASG grace period via props
+    const cfnAsg = pattern.autoScalingGroup.node.defaultChild as CfnAutoScalingGroup;
+    cfnAsg.healthCheckGracePeriod = 800;
+  }
+}
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -3,12 +3,18 @@ import "source-map-support/register";
 import { App } from "@aws-cdk/core";
 import { PrismStack } from "../lib/prism";
 import { PrismAccess } from "../lib/prism-access";
+import { PrismEc2App } from "../lib/prism-ec2-app";
 
 const app = new App();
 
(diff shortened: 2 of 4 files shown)


cdk/lib/prism-ec2-app.ts
```diff
@@ -0,0 +1,80 @@
...
+    // The pattern does not currently offer support for customising healthchecks via props
+    pattern.targetGroup.configureHealthCheck({
+      path: "/management/healthcheck",
+      unhealthyThresholdCount: 10,
+      interval: Duration.seconds(5),
+      timeout: Duration.seconds(3),
+    });
+
+    // Similarly the pattern does not offer support for extending the default ASG grace period via props
+    const cfnAsg = pattern.autoScalingGroup.node.defaultChild as CfnAutoScalingGroup;
```
jacobwinch, line 77: This ensures that we keep: https://github.com/guardian/prism/blob/b807073d2a2fad270b5576551772e46b81fd9fa5/cdk/lib/prism.ts#L95-L97

cdk/lib/prism-ec2-app.ts
```diff
@@ -0,0 +1,80 @@
...
+          deviceName: "/dev/sda1",
+          volume: BlockDeviceVolume.ebs(8, {
+            volumeType: EbsDeviceVolumeType.GP2,
+          }),
+        },
+      ],
+    });
+
+    // The pattern does not currently offer support for customising healthchecks via props
+    pattern.targetGroup.configureHealthCheck({
```
jacobwinch, line 69: This ensures that we keep:

https://github.com/guardian/prism/blob/b807073d2a2fad270b5576551772e46b81fd9fa5/cdk/lib/prism.ts#L115-L120

Which differs from [our default](https://github.com/guardian/cdk/blob/main/src/constructs/loadbalancing/alb/application-target-group.ts#L37-L44) significantly.

cdk/lib/prism-ec2-app.ts
```diff
@@ -0,0 +1,80 @@
...
+        },
+      },
+      certificateProps: {
+        CODE: { domainName: "prism.code.dev-gutools.co.uk" },
+        PROD: { domainName: "prism.gutools.co.uk" },
+      },
+      monitoringConfiguration: { noMonitoring: true },
+      access: { scope: AccessScope.INTERNAL, cidrRanges: [Peer.ipv4("10.0.0.0/8")] },
+      roleConfiguration: {
+        additionalPolicies: [
```
jacobwinch, line 40: This is just a copy+paste from: https://github.com/guardian/prism/blob/b807073d2a2fad270b5576551772e46b81fd9fa5/cdk/lib/prism.ts#L41-L53

cdk/lib/prism-ec2-app.ts
```diff
@@ -0,0 +1,80 @@
...
+    app: "prism",
+  };
+
+  constructor(scope: App, id: string, props: GuStackProps) {
+    super(scope, id, props);
+
+    const pattern = new GuPlayApp(this, {
+      ...PrismEc2App.app,
+      userData: {
+        distributable: {
```
jacobwinch, line 28: This is just a clone of:

https://github.com/guardian/prism/blob/b807073d2a2fad270b5576551772e46b81fd9fa5/cdk/lib/prism.ts#L70-L73

cdk/lib/prism-ec2-app.ts
```diff
@@ -0,0 +1,80 @@
...
+          new GuAssumeRolePolicy(this, "CrawlerPolicy", {
+            resources: ["arn:aws:iam::*:role/*Prism*", "arn:aws:iam::*:role/*prism*"],
+          }),
+        ],
+      },
+      scaling: {
+        CODE: { minimumInstances: 1 },
+        PROD: { minimumInstances: 2 },
+      },
+      blockDevices: [
```
jacobwinch, line 58: Similarly, this is lifted directly from:

https://github.com/guardian/prism/blob/b807073d2a2fad270b5576551772e46b81fd9fa5/cdk/lib/prism.ts#L99-L106

cdk/lib/prism-ec2-app.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const pattern = new GuPlayApp(this, {
+      ...PrismEc2App.app,
+      userData: {
+        distributable: {
+          fileName: "prism.deb",
+          executionStatement: `dpkg -i /${PrismEc2App.app.app}/prism.deb`,
+        },
+      },
+      certificateProps: {
```
jacobwinch, line 33: We will now create our own cert within the template, instead of passing in the arn via a parameter. We should be able to remove an unused cert when the migration process is complete.

cdk/lib/prism-ec2-app.ts
akash1810: Would be good to update the pattern to natively support custom healthchecks. One for later, I guess.

akash1810 (approved):

akash1810:
> I spent a long time experimenting with different migration strategies, but was not able to get them working with Prism

Is it worth writing the experience up anywhere for future reference/future tooling?

cdk/lib/prism-ec2-app.ts
jacobwinch: Yes, possibly. It's hard to tell how much of the variation here is due to careful tuning vs historical guesswork though!

jacobwinch:
> Is it worth writing the experience up anywhere for future reference/future tooling?

Fortunately I think that most of the learnings (read: problems!) are not widely applicable (since migrating from `@guardian/cdk` constructs to `@guardian/cdk` patterns is not a common path), but I will try to make a few notes for future reference.

cdk/lib/prism-ec2-app.ts
jacobwinch: Unused certificates have now been removed manually.

PR #246, opened 2021-07-09
Add an alarm to detect cycling instances

jacobwinch:
## What does this change?

Now that we are using the `GuPlayApp` pattern (https://github.com/guardian/prism/pull/245), we can easily add an alarm to detect cycling instances. This addresses one of the retro actions described [here](https://docs.google.com/document/d/11QMKU1J9C4aZR6cse_cE6vI1f_IKr0mLuALuwMpfJ3M/edit#heading=h.ofhrsl2zekxj). 

## How to test

I've deployed this to `CODE`. The alarm configuration itself was tested as part of https://github.com/guardian/cdk/pull/594.

We will be notified if Prism instances are frequently becoming unhealthy.

We should keep an eye on this alarm and tune or remove it if it generates too much noise.

### DIFF
--- a/cdk/lib/prism-ec2-app.ts
+++ b/cdk/lib/prism-ec2-app.ts
@@ -34,7 +34,11 @@ export class PrismEc2App extends GuStack {
         CODE: { domainName: "prism.code.dev-gutools.co.uk" },
         PROD: { domainName: "prism.gutools.co.uk" },
       },
-      monitoringConfiguration: { noMonitoring: true },
+      monitoringConfiguration: {
+        snsTopicName: "devx-alerts",
+        http5xxAlarm: false,
+        unhealthyInstancesAlarm: true,
+      },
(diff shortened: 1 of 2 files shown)


cdk/lib/prism-ec2-app.ts
```diff
@@ -34,7 +34,11 @@ export class PrismEc2App extends GuStack {
         CODE: { domainName: "prism.code.dev-gutools.co.uk" },
         PROD: { domainName: "prism.gutools.co.uk" },
       },
-      monitoringConfiguration: { noMonitoring: true },
+      monitoringConfiguration: {
+        snsTopicName: "devx-alerts",
+        http5xxAlarm: false,
```
jacobwinch, line 39: We should also consider enabling this in a subsequent PR.

akash1810 (approved):
Brilliant!

PR #259, opened 2021-09-20
chore: Use GitHub Actions for CI

akash1810:
## What does this change?

Switches CI from TeamCity to GitHub Actions (with the use of https://github.com/guardian/actions-assume-aws-role).

There are a few benefits to using GHA individually:
- No build queue
- No need to perform a dance with `nvm`
- No need to globally install `yarn`
- Build logs are co-located with the PR (via the `Checks` tab)

Collectively, use of GHA means we don't have to maintain our own CI infrastructure.

## How to test

`devx::prism` should appear as a deployable project in Riff-Raff. It should have two files: the cloudformation and the deb and it should be successfully deployed to CODE.

Deployment gets a little tricky as the Riff-Raff project name has changed and we all have to remember this. Riff-Raff restrictions help here though.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
+name: CI
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
jacobwinch, line 7: Ah, thanks for remembering this!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: guardian/actions-assume-aws-role@v1.0.0
+        with:
+          awsRoleToAssume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
+      - name: Use Node.js 14.15.5
+        uses: actions/setup-node@v2.4.0
+        with:
+          node-version: 14.15.5
+      - name: Set up JDK 11
```
jacobwinch, line 23: It looks like we are [still using Java 8 in TeamCity](https://teamcity.gutools.co.uk/admin/editBuildParams.html?id=buildType:Tools_Prism)(?), so perhaps we should stick with that for the initial migration:

image

Aside: it wasn't that easy to figure this out in TeamCity, hooray for moving this config into git!

build.sbt
```diff
@@ -21,7 +21,11 @@ lazy val root = (project in file("."))
     fileDescriptorLimit := Some("16384"),
     maintainer := "Guardian Developers <dig.dev.software@theguardian.com>",
     topLevelDirectory in Universal := Some(normalizedName.value),
+    riffRaffPackageName := s"devx::${name.value}",
+    riffRaffManifestProjectName := riffRaffPackageName.value,
     riffRaffPackageType := (packageBin in Debian).value,
+    riffRaffUploadArtifactBucket := Option("riffraff-artifact"),
+    riffRaffUploadManifestBucket := Option("riffraff-builds"),
```
akash1810, line 28: It would be nice if the plugin would default these buckets for the `riffRaffUpload` task (they don't seem to be needed for the `riffRaffNotifyTeamcity` task...)

.github/workflows/ci.yml
akash1810: Whoa! Great spot! Updated.

> Aside: it wasn't that easy to figure this out in TeamCity, hooray for moving this config into git!

🎉

sihil (approved):
This is great. I agree that we should default the buckets in the plugin. Do you want to do that before telling everyone?

build.sbt
akash1810: https://github.com/guardian/sbt-riffraff-artifact/pull/63

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v2
+      - uses: guardian/actions-assume-aws-role@v1.0.0
+        with:
+          awsRoleToAssume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
+      - name: Use Node.js 14.15.5
+        uses: actions/setup-node@v2.4.0
```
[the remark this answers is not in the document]
akash1810: Hmm, `guardian/actions-setup-node` isn't doing quite what I'd expect. The `.nvmrc` file contains `14.15.5`, however the action is using `v14.17.6`.

From https://github.com/guardian/prism/runs/3731913592?check_suite_focus=true 

image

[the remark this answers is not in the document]
sihil: 🤔  @sndrs ?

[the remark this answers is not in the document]
sihil: Perhaps this is because it's in a subdirectory. I guess that would make sense as you can't have multiple versions of node setup.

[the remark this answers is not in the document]
akash1810: This is fixed now. The issue was the v2.4.0 tag wasn't at the correct point in the history. We've created v2.4.1 to fix this. Now I need to work out how to configure caching 😄 .

akash1810:
This PR is almost at the end! Just waiting for Maven to show the latest release of `sbt-riffraff-plugin`.

sihil (approved):
🎉

PR #269, opened 2021-10-06
update prism to use @guardian/cdk 26.2.1

michaelwmcnamara:
## What does this change?
This change updates @guardian/cdk dependency to 26.2.1 with requisite changes to keep everything working correctly.

## How to test
run /script/test and fix any errors
Deploy to CODE

keeping things up-to-date

this change mitigates risks of cdk version becoming too far out of date. Main risk is that there is some unknown change in the latest versions of cdk which we havent captured in our testing, the greater risk is not to update though.


jacobwinch:
Thanks for updating this @michaelwmcnamara 👍

PR #282, opened 2021-12-13
Update to sbt version 1.5.6 and Play version 2.8.8

jacobwinch:
## What does this change?

This PR updates `sbt` to version `1.5.6`. 

This version of `sbt` was incompatible with Play `2.8.2`, so I have also bumped the Play version to `2.8.8` and made a few small changes to keep things working.

## How to test

I've confirmed that this change can be [successfully deployed to `CODE`](https://riffraff.gutools.co.uk/deployment/view/10f12972-ca84-4d67-acff-ce9aa50caba7).

* All functionality should work as before
* The `sbt` [version](https://github.com/sbt/sbt/releases/tag/v1.5.6) is no longer vulnerable to [Remote code injection in Log4j](https://github.com/advisories/GHSA-jfh8-c2jp-5v3q)

This should be pretty low risk due to testing described above.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -6,7 +6,7 @@ resolvers += "Typesafe repository" at "https://repo.typesafe.com/typesafe/releas
 
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
 
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.2")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.8")
 
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.10.0")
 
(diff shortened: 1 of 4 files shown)


project/plugins.sbt
```diff
@@ -6,7 +6,7 @@ resolvers += "Typesafe repository" at "https://repo.typesafe.com/typesafe/releas
 
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
 
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.2")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.8")
```
jacobwinch, line 9: It seems like this is the first version that's compatible with sbt `1.5.x` - see https://github.com/guardian/ophan/pull/4085 for more details.

jorgeazevedo (approved):

PR #291, opened 2022-02-02
chore: Bump `@guardian/cdk` to 34.0.0

akash1810:
## What does this change?

Bump GuCDK to the latest version.

### DIFF
--- a/cdk/lib/__snapshots__/prism-ec2-app.test.ts.snap
+++ b/cdk/lib/__snapshots__/prism-ec2-app.test.ts.snap
@@ -3,7 +3,7 @@
 exports[`The PrismEc2App stack matches the snapshot 1`] = `
 Object {
   "Mappings": Object {
-    "stagemapping": Object {
+    "prism": Object {
       "CODE": Object {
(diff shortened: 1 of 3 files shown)


cdk/lib/__snapshots__/prism-ec2-app.test.ts.snap
```diff
@@ -906,6 +906,10 @@ dpkg -i /prism/prism.deb",
             "Key": "deregistration_delay.timeout_seconds",
             "Value": "30",
           },
+          Object {
+            "Key": "stickiness.enabled",
+            "Value": "false",
```
akash1810, line 911: `false` was the implicit [default](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-targetgroup-targetgroupattribute.html) previously - this change should be a no-op.

jacobwinch (approved):
That diff is not too bad considering the fact that we're going from `26.2.1` to `34.0.0`!

PR #292, opened 2022-02-02
feat: Enable access logging

akash1810:
Requires #291.

## What does this change?

We want to understand exactly how RIff-Raff interacts with Prism to, ultimately, remove as much of the coupling as possible. This follows an issue where RIff-Raff 503'd because it struggled to process a response from Prism (https://github.com/guardian/riff-raff/pull/675).

Although the SSM parameter on `/account/services/access-logging/bucket` exists, the bucket it refers to does not 😱 ! What is the best way to create this bucket with IaC?

### DIFF
--- a/cdk/lib/__snapshots__/prism-ec2-app.test.ts.snap
+++ b/cdk/lib/__snapshots__/prism-ec2-app.test.ts.snap
@@ -34,6 +34,11 @@ Object {
       "Description": "Amazon Machine Image ID for the app prism. Use this in conjunction with AMIgo to keep AMIs up to date.",
       "Type": "AWS::EC2::Image::Id",
     },
+    "AccessLoggingBucket": Object {
+      "Default": "/account/services/access-logging/bucket",
+      "Description": "S3 bucket to store your access logs",
+      "Type": "AWS::SSM::Parameter::Value<String>",
+    },
     "DistributionBucketName": Object {
       "Default": "/account/services/artifact.bucket",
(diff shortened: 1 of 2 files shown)


cdk/lib/__snapshots__/prism-ec2-app.test.ts.snap
```diff
@@ -693,6 +680,20 @@ dpkg -i /prism/prism.deb",
...
+          },
+          Object {
+            "Key": "access_logs.s3.bucket",
+            "Value": Object {
+              "Ref": "AccessLoggingBucket",
+            },
+          },
+          Object {
+            "Key": "access_logs.s3.prefix",
+            "Value": "",
```
jacobwinch, line 695: Probably worth adding some kind of prefix (maybe just `prism`?) as I think this bucket is intended to be shared by multiple services

jacobwinch:
> What is the best way to create this bucket with IaC?

This is a great question. IMO we should only keep ELB access logs for 2 weeks (as they may contain personal data), which makes it even more important to define these buckets in a standardised way... 

Long-term solution: ultimately I think the 'account boilerplate' template which gives you a VPC, logging Kinesis stream, artifact bucket etc. will be the answer to this. 

Short-term solution: I wonder if we can just provide a `cdk` example of how to create this 'correctly' and add ours via `deploy-tools-platform`?

jacobwinch (approved):

PR #303, opened 2022-03-28
Add some missing stack owners

jacobwinch:
## What does this change?

This PR adds some missing stack owners. This will help more alerts to go to the right place and also helps us to better understand the number of apps that each team needs to migrate to GuCDK.

## More context

Prism has a concept of stack ownership. I'm not convinced that this the 'correct' place to track stack ownership moving forwards, but this is currently used to [help Amiable to send notifications about out of date AMIs](https://github.com/guardian/amiable/blob/4d00b110de22593ed869a583c1678c95d5af59f6/app/services/notification/ScheduledNotificationRunner.scala#L22) and is also used for tracking GuCDK adoption (https://github.com/guardian/prism/pull/300). 

If an owner for a stack is not found, DevX is used as a default. This means that we currently get lots of alerts about stacks that we don't maintain. It also makes it harder to plan/understand the migration effort required for each team when analysing GuCDK adoption data.

This list is still not 100% complete, but this PR should remove a lot of the noise.


app/data/Owners.scala
```diff
@@ -27,46 +27,55 @@ trait Owners {
 
 object Owners extends Owners {
 
-  override def default: Owner = Owner("dig.dev.tooling")
+  override def default: Owner = Owner("devx.sec.ops")
```
jacobwinch, line 30: I've replaced `dig.dev.tooling` with `devx.sec.ops` as the former also contains CSTI (and I don't think that they maintain many of the stacks that this ownership concept relates to)

app/data/Owners.scala
```diff
@@ -27,46 +27,55 @@ trait Owners {
...
     "digitalcms.dev" -> SSA("cms-fronts"),
     "digitalcms.dev" -> SSA("elk-new"),
+    "digitalcms.dev" -> SSA("media-service"),
+    "digitalcms.dev" -> SSA("grid-elasticsearch"),
+    "digitalcms.dev" -> SSA("composer"),
+    "digital.investigations" -> SSA("pfi"),
     "digital.investigations" -> SSA("pfi-giant"),
     "digital.investigations" -> SSA("pfi-playground"),
     "digital.investigations" -> SSA("pfi-structured"),
-    "thegrid.dev" -> SSA("media-service"),
```
jacobwinch, line 49: `thegrid.dev` also includes Central Production, who probably don't want alerts about AWS infrastructure, so we now use `digitalcms.dev` instead

app/data/Owners.scala
```diff
@@ -27,46 +27,55 @@ trait Owners {
...
     "content.platforms" -> SSA(stack = "pubflow"),
     "content.platforms" -> SSA(stack = "pubflow-preview"),
+    "mobile.server.side" -> SSA(stack = "content-api-mobile-notifications"),
     "mobile.server.side" -> SSA(stack = "mobile"),
     "mobile.server.side" -> SSA(stack = "mobile-notifications"),
     "mobile.server.side" -> SSA(stack = "mobile-preview"),
     "mobile.server.side" -> SSA(stack = "mobile-purchases"),
     "infosec.ops" -> SSA(stack = "infosec"),
-    "infosec.ops" -> SSA(stack = "nessus"),
+    "infosec.ops" -> SSA(stack = "infosec-nessus"),
```
jacobwinch, line 78: A quick GitHub search suggests that this was a typo

app/data/Owners.scala
```diff
@@ -27,46 +27,55 @@ trait Owners {
...
     "commercial.dev" -> SSA(stack = "frontend", app = Some("ipad-ad-preview")),
     "commercial.dev" -> SSA(stack = "flexible", app = Some("campaign-central")),
     "content.platforms" -> SSA(stack = "content-api"),
     "content.platforms" -> SSA(stack = "content-api-crier-v2"),
     "content.platforms" -> SSA(stack = "content-api-crier-v2-preview"),
     "content.platforms" -> SSA(stack = "content-api-dashboard"),
     "content.platforms" -> SSA(stack = "content-api-facebook-news-bot"),
+    "content.platforms" -> SSA(stack = "content-api-facebook-newstab"),
+    "content.platforms" -> SSA(stack = "content-api-fastly-cache-purger"),
+    "content.platforms" -> SSA(stack = "content-api-floodgate"),
```
akash1810, line 63: minor: wonder if there's benefit to supporting wildcarding, e.g. `content-api-*`?

akash1810 (approved):

app/data/Owners.scala
jacobwinch: Yeah, I think that would definitely help to stop this list from drifting out of date so easily

PR #334, opened 2022-07-19
Re-enable cloudformation crawling

nicl:
This collector was introduced [here](https://github.com/guardian/prism/pull/197).

But [quickly disabled](https://github.com/guardian/prism/pull/228) as it was hitting rate limits.

The feature would be useful for some CDK tracking we are doing andI wonder if using the 'slow' (once per day) crawl rate will help here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,11 @@ dist
 Gemfile.lock
 .rakeTasks
 *.iml
+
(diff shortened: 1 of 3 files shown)


jacobwinch (approved):
Let's give this another try 👍

PR #340, opened 2022-08-10
Add running locally instructions to README

kenoir:
## What does this change?

This change adds some documentation about how to run locally in order that this can be done  by folk new to the project.

We remove the outdated `./sbt` script from the root that is no longer used (and indicates using a misleading version of Java). A `.java-version` file is added to help any version management tools (or humans) to identify the correct version of Java to use.

## How to test

Follow the instructions, do they work?

Folk can run the app locally easily.


README.md
```diff
@@ -67,3 +67,18 @@ A typical Prism response includes the following top level fields:
...
 There is a [ruby gem](./marauder/README.md) that allows you to query the api from the command line.
+
+
+Development
+------------------------------
+1. Get Security Janus credentials.
+
+2. Ensure you are running Java 11
+
+3. Update the `play.http.secret.key` configuration in `conf/application.conf` to a dummy value
```
jfsoul, line 78: Additionally: create the gc log file if it doesn't already exist?

jfsoul: Also, thinking about it, why doesn't the `application.conf` have a dummy value already?

jfsoul (approved):

README.md
kenoir: Turns out these are not required if you start the app using `run` instead of `start`!

jfsoul (approved):

PR #386, opened 2022-09-28
Add subnet scope to VPC reports

nicl:
A subnet is considered private unless it has an internet gateway associated to its route table.

This is helpful for us to better understand the shape of VPCs across the dept - particularly the number of VPCs that do not have any private subnets.

### DIFF
--- a/app/collectors/vpc.scala
+++ b/app/collectors/vpc.scala
@@ -11,6 +11,7 @@ import utils.Logging
 import scala.jdk.CollectionConverters._
 import scala.language.postfixOps
 import scala.util.Try
+import software.amazon.awssdk.services.ec2.model.DescribeRouteTablesRequest
 
 class VpcCollectorSet(accounts: Accounts) extends CollectorSet[Vpc](ResourceType("Vpc"), accounts, Some(Regional)) {
   val lookupCollector: PartialFunction[Origin, Collector[Vpc]] = {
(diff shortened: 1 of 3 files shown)


kenoir (approved):
👌

app/collectors/vpc.scala
```diff
@@ -27,17 +28,46 @@ case class AWSVpcCollector(origin:AmazonOrigin, resource: ResourceType, crawlRat
...
+      // internet gateways, or nats.
+      val assocs = table.associations().asScala
+
+      val subnetInfo = for {
+        // Discard route tables that do not have an explicit subnet association.
+        subnetAssoc <- assocs.find(a => a.subnetId().nonEmpty)
+
+        // Public subnets have route tables with an associated internet gateway.
+        gatewayAssoc <- assocs.find(a => a.gatewayId().nonEmpty)
+        scope = if (gatewayAssoc.gatewayId().startsWith("igw")) Public else Private
```
akash1810, line 51: Worth adding an inline comment here? `startsWith("igw")` can't be the recommended way to detect this?! 😅 🤯

akash1810 (approved):

app/collectors/vpc.scala
nicl: Yeah it's a strange one. Have added a comment to explain. Feel like there must be a cleaner way too but can't find it.

akash1810 (dismissed):

app/collectors/vpc.scala
```diff
@@ -27,17 +29,55 @@ case class AWSVpcCollector(origin:AmazonOrigin, resource: ResourceType, crawlRat
...
+
+      // It feels like there should be a better way to detect the presence of an AWS Internet Gateway but apparently this is it :(.
+      val tableHasIgw = routes.exists(route => Option(route.gatewayId()).getOrElse("").startsWith("igw"))
+      val subnetIDs = assocs.flatMap(assoc => Option(assoc.subnetId()).toList)
+      RouteTable(isMain = isMain, hasInternetGateway = tableHasIgw, subnetIDs = subnetIDs.toSet)
+    })
+
+    val data = subnets.map(subnet => {
+      // If there is no explicit route table associated with a subnet, the VPC 'main' route table is used instead.
+      val main = tables.find(table => table.isMain)
```
akash1810, line 57: Worth linking to any AWS docs here?

akash1810 (approved):

PR #387, opened 2022-09-29
Add VPC isDefault

nicl:
It's useful to know which VPCs are default as:

1) we don't want people to use these
2) we can ignore them in analysis e.g. of VPCs that do not have private subnets (which default VPCs lack)

See:

* https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html
* https://github.com/guardian/recommendations/blob/main/AWS.md#vpc


jacobwinch (approved):

nicl:
Somehow I missed that this was already present. Thanks @jacobwinch for the tip :)

PR #408, opened 2022-10-12
Use getCallerIdentity instead of getUser

JuliaBrigitte:
Use getCallerIdentity instead of getUser to enable better results while running locally (because of a permission problem).

Co-authored-by Akash and Nic

## What does this change?

Switching form getUser to getCallerIdentity will ensure accountNumber will show when running locally because of an permission problem with aws credentials preventing this

## How to test

http://localhost:9000/sources/accounts
Should not return ???????? for accountNumber any more

Any side effects would only result in problems within devX, so we feel we can go ahead.

### DIFF
--- a/app/agent/origin.scala
+++ b/app/agent/origin.scala
@@ -23,13 +23,11 @@ import scala.util.control.NonFatal
 import scala.util.matching.Regex
 
 class Accounts(prismConfiguration: PrismConfiguration) extends Logging {
-  val ArnIamAccountExtractor: Regex = """arn:aws:iam::(\d+):user.*""".r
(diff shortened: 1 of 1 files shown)


app/agent/origin.scala
```diff
@@ -26,13 +26,12 @@ class Accounts(prismConfiguration: PrismConfiguration) extends Logging {
   val ArnIamAccountExtractor: Regex = """arn:aws:iam::(\d+):user.*""".r
```
akash1810, line 26: `ArnIamAccountExtractor` is unused now, shall we remove?

akash1810 (dismissed):

app/agent/origin.scala
JuliaBrigitte: Thank, yes I did, missed that when I split the original PR in two.

akash1810 (approved):

PR #410, opened 2022-10-14
feat: Add a request logger

akash1810:
A variant of https://github.com/guardian/prism/pull/292.

## What does this change?

It will be useful to know how Prism is accessed. In this change, we add a filter to log each request. The log lines have some markers, so we can interrogate them more easily in Central ELK.

## How to test

- Deploy to CODE
- Interact with Prism CODE
- View [logs in Central ELK](https://logs.gutools.co.uk/s/devx/goto/29072010-4bb5-11ed-a02a-2d44512a4a41)

We know which endpoints are in use.

N/A

## Images

image

### DIFF
--- a/app/RequestLoggingFilter.scala
+++ b/app/RequestLoggingFilter.scala
@@ -1,0 +1,67 @@
+import akka.stream.Materializer
+import net.logstash.logback.marker.Markers.appendEntries
+import play.api.mvc.{Filter, RequestHeader, Result}
+import play.api.MarkerContext
+import utils.Logging
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+import scala.util.{Failure, Success}
+
+class RequestLoggingFilter(override val mat: Materializer)(implicit ec: ExecutionContext) extends Filter with Logging {
+  override def apply(next: (RequestHeader) => Future[Result])(rh: RequestHeader): Future[Result] = {
+    val start = System.currentTimeMillis()
+    val result = next(rh)
+
+    result onComplete {
+      case Success(response) =>
+        val duration = System.currentTimeMillis() - start
+        logSuccess(rh, response, duration)
+
+      case Failure(err) =>
+        val duration = System.currentTimeMillis() - start
+        logFailure(rh, err, duration)
+    }
+
+    result
+  }
+
+  private def logSuccess(request: RequestHeader, response: Result, duration: Long): Unit = {
+    val originIp = request.headers.get("X-Forwarded-For").getOrElse(request.remoteAddress)
+    val referer = request.headers.get("Referer").getOrElse("")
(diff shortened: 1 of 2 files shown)


app/RequestLoggingFilter.scala
```diff
@@ -0,0 +1,67 @@
...
+        logFailure(rh, err, duration)
+    }
+
+    result
+  }
+
+  private def logSuccess(request: RequestHeader, response: Result, duration: Long): Unit = {
+    val originIp = request.headers.get("X-Forwarded-For").getOrElse(request.remoteAddress)
+    val referer = request.headers.get("Referer").getOrElse("")
+    val userAgent = request.headers.get("User-Agent").getOrElse("")
```
akash1810, line 32: Can be used to distinguish human vs. machine users. For example, a user-agent of "node-fetch" is a machine.

app/RequestLoggingFilter.scala
```diff
@@ -0,0 +1,67 @@
...
+      "duration" -> duration,
+      "requestUri" -> request.uri,
+      "userAgent" -> userAgent
+    )
+
+    val markers = MarkerContext(appendEntries(mandatoryMarkers.asJava))
+    log.info(s"""$originIp - "${request.method} ${request.uri} ${request.version}" ${response.header.status} $length "$referer" ${duration}ms""")(markers)
+  }
+
+  private def logFailure(request: RequestHeader, throwable: Throwable, duration: Long): Unit = {
```
kenoir, line 49: Probably not quite worth DRYing this up yet.

app/RequestLoggingFilter.scala
```diff
@@ -0,0 +1,67 @@
...
+import net.logstash.logback.marker.Markers.appendEntries
+import play.api.mvc.{Filter, RequestHeader, Result}
+import play.api.MarkerContext
+import utils.Logging
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+import scala.util.{Failure, Success}
+
+class RequestLoggingFilter(override val mat: Materializer)(implicit ec: ExecutionContext) extends Filter with Logging {
```
kenoir, line 11: I'm surprised it's required to implement this ourselves and there isn't some existing solution to produce access logs. Is it possible to get these from the ELB?

I'm presuming that you've thought of this and there is a good reason, so approving, but interested in the thought process.

kenoir (approved):

app/RequestLoggingFilter.scala
akash1810: https://github.com/guardian/prism/pull/292 does the ELB based logging, however there are a couple of gaps there:
  - Where to define the bucket (aws-account-setup might be a good fit)
  - How to query the logs, as we don't have a route from S3 to Central ELK at the moment

PR #472, opened 2023-01-02
Introduce Scalafmt

nicl:
We are migrating all DevX Scala tooling to Scalafmt. See e.g. https://github.com/guardian/riff-raff/pull/940 and also this Trello card: https://trello.com/c/eIgRRUgB/1634-introduce-scalafmt-to-our-scala-repos.


scripts/ci
```diff
@@ -13,4 +13,4 @@ fi
   ./script/ci
 )
 
-sbt clean compile test riffRaffUpload
+sbt clean scalafmtCheckAll compile test riffRaffUpload
```
kelvin-chappell, line 16: Less important but can also include `scalafmtSbtCheck` for SBT config.
https://scalameta.org/scalafmt/docs/installation.html#task-keys

nicl: Ah great shout, hadn't noticed that wasn't covered. Will do.

akash1810 (approved):

PR #530, opened 2023-02-15
Use arm64-focal-java11-deploy-infrastructure as the AMI

kenoir:
## What does this change?

Move to use https://amigo.gutools.co.uk/recipes/arm64-focal-java11-deploy-infrastructure.

This will allow us to:

- Move Prism to 20.04 to avoid the 18.04 EOL
- Test whether the new `arm64-focal-java11-deploy-infrastructure` image is working as expected

## How to test

Deploy this PR to the CODE environment and check the EC2 instances with the new AMI are successfully brought into service & that we can access the Prism API.

See: [TBD]

Can we use the focal AMI to successfully deploy Prism?

Prism is not deployable with Focal base AMIs, we'll revert.


NovemberTang (approved):

kenoir:
Tested this by accessing the instance and curling locally to ensure the API has come up.

PR #790, opened 2023-11-09
Upgrade to play 3.0.0

JuliaBrigitte:
The dependency PR upgrading to play 3.0.0 fails because you have to migrate akka to pekko manually
[Doc Migration to play 3.0. akka to pekko](https://medium.com/xworks-techinsights/akka-to-pekko-migration-step-by-step-cca64bc61016)

## What does this change?

Upgrade to play 3.0.0

## How to test

Deployed to and checked on code (https://prism.code.dev-gutools.co.uk/)

### DIFF
--- a/app/RequestLoggingFilter.scala
+++ b/app/RequestLoggingFilter.scala
@@ -1,4 +1,4 @@
-import akka.stream.Materializer
+import org.apache.pekko.stream.Materializer
 import net.logstash.logback.marker.Markers.appendEntries
 import play.api.mvc.{Filter, RequestHeader, Result}
 import play.api.MarkerContext
--- a/app/agent/collector.scala
+++ b/app/agent/collector.scala
@@ -1,9 +1,9 @@
 package agent
 
-import akka.actor.ActorSystem
-import akka.agent.Agent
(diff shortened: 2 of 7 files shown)


akash1810 (commented):
This looks great! 🏆 

Do we need to change https://github.com/guardian/prism/blob/ee2f2c6276d8a112d1449276ee68d1b5985d95c5/conf/application.conf#L24 too? Prism is fairly chatty in it's logs, so we should be able to confirm in Central ELK?

AshCorr (commented):
Awesome!

We should be able to remove [this repository resolver](https://github.com/guardian/prism/pull/790/files#diff-8fd74420ca55aa91d4c53775d2a55a999c405b5a4fafeaf73b7f0ef83e1f3b0cR5) now, Play 3.0 can be found on Central maven

Looks like this dependency is still on Play 2.x https://github.com/guardian/prism/blob/main/build.sbt#L70 , because the play groupID has changed in 3.x I think this might cause 2 different versions of Play to be bundled into our class path... not quite sure what the ramifications of that is 🤔

akash1810 (approved):
img

PR #875, opened 2024-04-16
Drop `play-json-extensions` dependency

AshCorr:
## What does this change?

Drops the `play-json-extensions` dependency. This dependency was used to auto-generate JSON Writer implementations for case classes with more than 22 fields.

Unfortunately it appears this dependency hasn't been updated in a long time and [doesn't fully support Play 3 yet](https://github.com/bizzabo/play-json-extensions/issues/94) causing us to have to do some trickery with dependencies to avoid loading in any Play 2 dependencies, specifically Akka.

Thankfully we only had 1 class that was using this extension which was quite easy to write a custom JSON Writer implementation for.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -59,7 +59,6 @@ lazy val root = (project in file("."))
       "com.amazonaws" % "aws-java-sdk-autoscaling" % awsVersionOne,
       "org.playframework" %% "play-json" % "3.0.2",
       "org.playframework" %% "play-json-joda" % "3.0.2",
-      "ai.x" %% "play-json-extensions" % "0.42.0",
       ws,
       "org.scala-stm" %% "scala-stm" % "0.11.1",
       filters,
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -69,10 +68,6 @@ lazy val root = (project in file("."))
       "ch.qos.logback" % "logback-classic" % "1.5.5",
       "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.17.0"
     ),
-    excludeDependencies ++= Seq(
-      // As of Play 3.0, groupId has changed to org.playframework; exclude transitive dependencies to the old artifacts
-      ExclusionRule(organization = "com.typesafe.play")
```
AshCorr, line 74: `play-json-extensions` was trying to bring in `com.typesafe.play:play-json`

akash1810 (approved):
Ah, brilliant!

PR #918, opened 2024-07-11
Add launch template support

philmcmahon:
## What does this change?
This adds a new resource type to prism - launch templates. Launch Templates are now used widely at the guardian, they are the default in patterns in the [guardian cdk project](https://github.com/guardian/cdk). It's helpful to have them in prism so that tools such as amigo can query prism to find out whether a particular AMI is in use (currently as part of its housekeeping job AMIgo will delete AMIs that are actually still in use because it doesn't check launch template usages).

Launch templates are a bit more complicated than launch configurations as they have multiple versions. Here I've decided that we only care about launch template versions that are in use by an autoscaling group.

## How to test
I've tested this on prism CODE (it's live now https://prism.code.dev-gutools.co.uk/launch-templates)


philmcmahon:
Note that no extra permissions are required for this change, as prism already has describe* for the ec2 and autoscaling namespaces https://github.com/guardian/prism/blob/main/cdk/lib/prism-access.ts#L51

app/collectors/launchTemplates.scala
```diff
@@ -0,0 +1,122 @@
...
+      // might need to filter out asgs which don't have launch templates?
+      val requestLt = DescribeLaunchTemplateVersionsRequest.builder
+        .launchTemplateId(lt.id)
+        .versions(lt.version)
+        .build
+      val template = ec2Client
+        .describeLaunchTemplateVersions(requestLt)
+        .launchTemplateVersions()
+        .asScala
+        .head
```
adamnfish, line 81: Are we sure we don't need to capture the other versions? I think it's rare to do versioned deployments at present, but I'm hoping that will start to change soon, maybe even this quarter! (Versioned deployments would unlock automated rollback of failed deplyments, for example)

Since we're motivated by deletinbg old AMIs in Amigo, this would create a decision about how far back we'd like our launch templates to still be launchable, but that's a decision for Amigo to make, rather than Prism. That suggests Prism should be exposing at least as much information as Amigo needs to make that decision, which would certainly more than just the most recent?

I may have misunderstood this bit, in which case apologies!

app/collectors/launchTemplates.scala
```diff
@@ -0,0 +1,122 @@
...
+    LaunchTemplateVersion(
+      arn = config.launchTemplateId(),
+      name = config.launchTemplateName(),
+      imageId = config.launchTemplateData().imageId(),
+      region = origin.region,
+      instanceProfile =
+        Option(config.launchTemplateData().iamInstanceProfile().arn()),
+      createdTime = Try(config.createTime()).toOption,
+      instanceType = config.launchTemplateData().instanceType().toString,
+      securityGroups = Option(
```
akash1810, line 102: It looks like these are all empty, is that expected?

```sh
curl --silent https://prism.code.dev-gutools.co.uk/launch-templates | \
jq ".data.\"asg-launch-templates\"[].securityGroups" | \
uniq
```

app/collectors/launchTemplates.scala
```diff
@@ -0,0 +1,122 @@
...
+        config.launchTemplateData().securityGroups()
+      ).toList.flatMap(_.asScala).map { sg =>
+        s"arn:aws:ec2:${origin.region}:${origin.accountNumber.get}:security-group/$sg"
+      }
+    )
+  }
+}
+
+case class LaunchTemplateVersion(
```
adamnfish, line 111: Should we be capturing the tags for the launch template as well? It looks like the tags are available on the launchTemplateData via the `tagSpecifications` field?

adamnfish (commented):
Great, thank you!

Couple of questions, chiefly about the versions. Let's chat about that, but basically LGTM.

app/collectors/launchTemplates.scala
philmcmahon: Hi again, I refactored this in https://github.com/guardian/prism/pull/918/commits/9392471a8d4590162cb9becde524f4f5d87d919f to fetch all versions of all launch templates. Unfortunately the deploy failed, I think because several of the launch template collectors repeatedely failed, and prism only passes the healthcheck once it has managed to fetch everything for each account at least once (see https://logs.gutools.co.uk/s/devx/goto/6c624950-44fd-11ef-8041-ff280ea02022)

philmcmahon: I'm not sure where to go next, but I am running out of time to work on this, so am tempted to suggest we stick with the original solution, unless you have any better ideas? For the deploy tools account alone there are 581 launch templates, so my guess is that there are quite a lot more for accounts like frontend. Prism should be able to handle this, but just means thinking a lot more about paging etc than we have had to before for this application.

app/collectors/launchTemplates.scala
```diff
@@ -0,0 +1,122 @@
...
+    .region(origin.awsRegionV2)
+    .overrideConfiguration(AWS.clientConfig)
+    .build
+
+  def crawl: Iterable[LaunchTemplateVersion] = {
+    val asgs =
+      asgClient.describeAutoScalingGroups().autoScalingGroups().asScala.toList
+    val launchTemplates = asgs
+      // not all asgs have launch templates (they might have launch configurations instead
+      .filter(asg => asg.launchTemplate() != null)
```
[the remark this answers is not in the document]
philmcmahon: sounds good to me, will refactor once we've resolved the below questions

app/collectors/launchTemplates.scala
philmcmahon: Sounds like a bug to me! Will have a look

app/collectors/launchTemplates.scala
philmcmahon: Hey @adamnfish  - what do you think we should do here?

app/collectors/launchTemplates.scala
philmcmahon: Resolved in https://github.com/guardian/prism/pull/918/commits/10ffc57a7094df7cb88e54b52b88e9bdfa57bc25

app/collectors/launchTemplates.scala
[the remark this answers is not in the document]
philmcmahon: Resolved in https://github.com/guardian/prism/pull/918/commits/ee1d12ac2e6cdf2c87dcda6cb6767184d5910a9e

app/collectors/launchTemplates.scala
philmcmahon: Unfortunately the tags on the launch template aren't available - we would have to make another API call. `tagSpecifications` is the tags that will be applied to instances launched using this launch template. I'm inclined to leave them out of prism for now until we need them...

adamnfish (approved):
@philmcmahon has pointed out that we expect the collection of launch templates to only have a single result, so calling `head` on it isn't throwing away valuable information! Apologies for misreading that.

LGTM, and I like the naming change.

PR #946, opened 2024-09-19
chore(lint): Fail on `prettier` errors

akash1810:
## What does this change?
Configure `eslint` to fail on `prettier` errors. This ensures a consistent style for quote marks, indent levels, etc. It also means the project uses the departmental code style.

This is a no-op, as demonstrated by the CDK snapshot remaining unchanged.


cdk/package.json
akash1810, line 1: For the most part, this file is generated by GuCDK. It should be adding `eslint-plugin-prettier`...

jacobwinch (approved):

PR #948, opened 2024-09-19
feat: Perform EC2 deployments via CloudFormation

akash1810:
## What does this change?
Uses the new `GuEc2AppExperimental` available in [GuCDK v59.5.0](https://github.com/guardian/cdk/releases/tag/v59.5.0) to deploy changes to the autoscaling group entirely via a CloudFormation update.

## How to test
Deploy the branch?! It should succeed. This has been [done on CODE](https://riffraff.gutools.co.uk/deployment/view/2d014f20-ed9b-4164-b986-dc017e90dc34).

More data points to evaluate the effectiveness of `GuEc2AppExperimental`.

### DIFF
--- a/cdk/lib/prism.ts
+++ b/cdk/lib/prism.ts
@@ -25,6 +25,12 @@ import {
 interface PrismProps extends Omit<GuStackProps, 'description' | 'stack'> {
 	domainName: string;
 	minimumInstances: number;
+
+	/**
+	 * Which application build to run.
+	 * This will typically match the build number provided by CI.
+	 */
+	buildIdentifier: string;
 }
 
 export class Prism extends GuStack {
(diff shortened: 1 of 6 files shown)


cdk/lib/prism.ts
```diff
@@ -37,7 +43,13 @@ export class Prism extends GuStack {
 			app,
 		});
 
-		const pattern = new GuPlayApp(this, {
+		const { buildIdentifier } = props;
+
+		const filename = `${app}-${buildIdentifier}.deb`;
```
adamnfish, line 48: The most minor thing in the world, but I get a niggle that we could use a different separator character here? Hyphens are commonly to separate words in the artefct name, making the version pop out will help at a glance, and if someone wants to parse it out for any reason later. I[t looks like](https://github.com/guardian/prism/pull/948/files#diff-752b7c938d22343dca2a196affbde2ca0132902bb317d08f9ad3329da1be341fR17) sbt-native-packager is using `_` as the version delimiter, could we do the same?

akash1810: Can do. Will raise a follow-up PR. Would you think we can formalise the use of `_`? I've been struggling to think how we'd address this TODO - https://github.com/guardian/cdk/blob/890c89e07d6715277ef9db4f01baeafddc7696e8/src/experimental/patterns/ec2-app.ts#L265.

PR #1011, opened 2025-04-16
chore: Remove ACM certificate collector 

akash1810:
## What does this change?
Originally added in https://github.com/guardian/prism/pull/47, Prism collected ACM certificate details to gain visibility of expiring certificates. Today, we're exclusively using DNS validated certificates which auto-renew. That is, having this data in Prism is no longer necessary.

Lastly, Service Catalogue can be used to understand ACM details today.


cdk/lib/prism-access.ts
akash1810, line 1: These changes would need manual deployment to the CloudFormation Stack Set.

akash1810:
This cannot be merged yet as these endpoints are still used by https://github.com/guardian/dns-validation-lambda.

PR #1051, opened 2025-10-08
feat: Only read configuration from disk

akash1810:
> [!NOTE]
> Easiest to review [commit by commit](https://github.com/guardian/prism/pull/1051/commits).

## What does this change?
We currently load configuration from [a few places](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/app/AppLoader.scala#L21-L37):
1. [`application.conf`](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/conf/application.conf)
2. [Three rows](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/app/conf/DynamoConfiguration.scala#L102-L104) in a DynamoDb table
3.  [`env/global.conf`](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/app/conf/FileConfiguration.scala#L14-L16) (which is [checked in](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/conf/env/global.conf))
4. [`env/<STAGE>.conf`](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/app/conf/FileConfiguration.scala#L20-L24)
5. [`$HOME/.gu/deploy-prism.conf`](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/app/conf/FileConfiguration.scala#L25-L30)

Once loaded, the configurations are [merged together and passed to `AppComponents`](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/app/AppLoader.scala#L36-L42). This is quite complex! Indeed, the files at 4 and 5 do not actually exist.

This change simplifies configuration by only loading from three locations on disk[^1]:
1. `application.conf`
2. `env/global.conf`
3. Conditional:
   - When running in AWS, `/etc/gu/prism/<STAGE>.conf`[^2]
   - When running In DEV, `$HOME/.gu/prism/DEV.conf`[^3]

This means `DynamoConfiguration.scala` is now unused and has been removed. Consequently the dependency `"com.amazonaws" % "aws-java-sdk-dynamodb"` is redundant and is also removed. Indeed, this was the last AWS SDK v1 dependency, which reaches [end-of-support o 31 December 2025](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/). We _could_ update to the v2 SDK. However the shape of the `getItem` response is vastly different and would need a [fair amount of massaging](https://github.com/guardian/support-frontend/blob/a7e1d88515e7716dd7b899ac5c9a8c42ab5af0dd/support-frontend/app/services/LandingPageTestService.scala#L34-L65) before it can be passed to `ConfigFactory`.

<details><summary>Further cleanup</summary>
<p>

The screenshot below shows the contents of the `config-deploy` DynamoDb table. It only includes configuration for Prism. After this change, the table should be unused and can be deleted. We'd need to use metrics and/or CloudTrail to check there is no activity first though.

image

</p>
</details> 

## How to test
In addition to listing AWS accounts to crawl, the DynamoDb config also contained the Play secret. Given [Play won't start in production mode without a secret](https://www.playframework.com/documentation/2.3.7/Production#The-application-secret), one test for this change is checking it deploys as the secret will now come from `/etc/gu/prism/<STAGE>.conf`. I can confirm [it does](https://riffraff.gutools.co.uk/deployment/view/0bec9903-f074-4d42-9989-7f7b396bbafa). This also means we don't have to stage these changes across multiple PRs; we can safely remove the DynamoDb logic immediately.

Another test is to check the same sources are being collected. For this I've:
1. [Deployed `main` to CODE](https://riffraff.gutools.co.uk/deployment/view/8a99999f-b17f-45a7-af3c-3a97b78b8524)
2. Checked the `/sources/accounts` response
3. [Deployed this branch to CODE](https://riffraff.gutools.co.uk/deployment/view/0bec9903-f074-4d42-9989-7f7b396bbafa)
4. Compared the `/sources/accounts` response to 2

   This endpoint has a date field, rather than comparing the entire response (as the dates will differ), we can compare just the `.data` section:

   ```bash
   # From main
   ❯ curl https://<Prism-CODE-domain>/sources/accounts --silent | jq .data | md5
   208ba7bfff1187634d9344ccfcfe8b2b

   # From this branch
   ❯ curl https://<Prism-CODE-domain>/sources/accounts --silent | jq .data | md5
   208ba7bfff1187634d9344ccfcfe8b2b
   ```

The [healthcheck](https://github.com/guardian/prism/blob/364979f10c3b8a0a3e748d7e9db47474ecf101b3/app/controllers/Api.scala#L131C7-L150) only passes if all sources have been crawled. As both versions of the app have the same sources, and pass their healthcheck, we can assert that this version is equivalent to `main`.

- Simplified configuration strategy
- Fewer dependencies

[^1]: This could be further simplified by using the [`include(</PATH/TO/FILE>)` syntax](https://www.playframework.com/documentation/3.0.x/ConfigFile#Specifying-an-alternative-configuration-file).
[^2]: This file is downloaded from S3 in the UserData. The S3 file was created via https://github.com/guardian/prism/pull/1054.
[^3]: This file is downloaded from S3 via a new `./scripts/setup` file.


jacobwinch (approved):
Great work!

akash1810:
> The screenshot below shows the contents of the `config-deploy` DynamoDb table. It only includes configuration for Prism. After this change, the table should be unused and can be deleted. We'd need to use metrics and/or CloudTrail to check there is no activity first though.

There's been no recent read activity:

image

The table was deleted today, 11 May 2026.

PR #1131, opened 2026-05-04
Revert "chore(deps): Patch updates"

JuliaBrigitte:
Reverts guardian/prism#1129 because of failed deploy

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.12.10
+sbt.version=1.12.9


akash1810:
It looks like the [deployment failed](https://riffraff.gutools.co.uk/deployment/view/fe4cc669-4558-41a7-a581-a987e4e87b4b) due to the new instance not reporting healthy in the [allotted time](https://github.com/guardian/prism/blob/a83426d7a7f9f42d454e35979f99dbcf2dbbba53/cdk/lib/prism.ts#L145-L151). Unfortunately, this is more common than we'd like; we haven't prioritised fixing this 😢. I've [successfully deployed the latest build of `main`](https://riffraff.gutools.co.uk/deployment/view/addaa3d0-1a9e-4ddb-80e8-f91d19b05ff4). I don't think this PR is needed.

